PR #14, opened 2018-03-29
Added RGB keycodes to the configurator.

FrostByte-Dev:
Tried keeping the keycode names as short as possible to keep a clean aesthetic on the ui.


jackhumbert:
Perfect - thanks!

PR #20, opened 2018-03-30
Create .editorconfig

jackhumbert:


.editorconfig
```diff
@@ -0,0 +1,32 @@
...
+
+[Makefile]
+end_of_line = lf
+
+[*.sh]
+end_of_line = lf
+
+[*.json]
+end_of_line = lf
+
```
yanfali: probably should add *.js and *.css

PR #21, opened 2018-03-30
Catch errors in keycodes

yanfali:
While loading a converted keymap that contains unknown codes,
 loading stops for the remaining keycodes and an error is printed
 on the console.

 This commit extracts the lookup into a function which prints a
 warning if the code is unknown.

 - print a warning if keyboard code is unknown
 - gracefully handle a lookup failure by continuing to process keyboard
   codes

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -64,10 +64,8 @@ $(document).ready(() => {
 
   function loadDefault() {
     // hard-coding planck as the only default right now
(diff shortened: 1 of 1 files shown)


yanfali:
Example of output

screen shot 2018-03-30 at 01 13 26

jackhumbert:
Nice! Thanks :)

PR #22, opened 2018-03-30
Extract jQuery references, surface unknown codes

yanfali:
- pull jquery DOM lookups into variables
 - use those references
 - add stats on loading default keyboard including a list of skipped
   keycodes


yanfali:
screen shot 2018-03-30 at 09 42 37

jackhumbert:
Awesome - thanks!

PR #25, opened 2018-03-30
Hid arrows on number input.

FrostByte-Dev:
I also used prettier on the css file to clean it up a bit.


jackhumbert:
Nice :) thanks!

PR #26, opened 2018-03-31
Extract keycode parsing from loop

yanfali:
- define newKey and newAnyKey helpers to create key objects
 - simplify parsing logic
   - use helpers instead of constructing object literals
   - return AnyKey objects instead of skipping when we don't know the keycode


jackhumbert:
Nice! Thanks :)

Issue #24, opened 2018-03-30
console output gets crowded/messy when switching between different keyboards


mechmerlin:
The current workaround is to do a page refresh.

Some suggestions
1. Include a button to clear the console output
2. Auto clear whenever a new board is selected
3. Improve error messages. Rather than "No default for this keyboard", specify which keyboard.

PR #27, opened 2018-03-31
Issue #24 Clearing Status between keyboards

yanfali:
- use $status.html() to clear DOM. $status.val() causes the DOM to stop
   updating.
 - also provide a slightly friendlier message when we don't have a
   default for a particular keyboard.


jackhumbert:
Nice :) thanks!

PR #31, opened 2018-04-01
Loading default keymap highlight layers

yanfali:
- add lodash as a script depedency
  - add classes to index the layers to markup
  - if a layer contains keycodes highlight it in the layer UI


jackhumbert:
Nice - thanks!

Issue #33, opened 2018-04-01
Reset 'active' layer to 0 on loading default


yanfali:
If you are in the layer UI, and click load default, the active layer stays on the layer you last clicked on. It should reset to 0.

reproduce:

1. select planck keyboard
2. click layer (10)
3. load default

Symption:

layer 10 is still active but you are actually looking at layer 0.

yanfali:
Also introduces extra layers empty layers which will cause problems for backend

yanfali:
we can close this I think @jackhumbert

jackhumbert:
Awesome - thanks :)

PR #34, opened 2018-04-01
Issue #33 reset active layer to 0

yanfali:
- fixes UI state being incorrect on Load Default or Import Layer
 - resets current layer to 0
 - prevents bug where additional empty layers can be created


jackhumbert:
Nice - thanks!

PR #35, opened 2018-04-01
Cache keycode lookups

yanfali:
The current keycode lookup code does a linear, O(n), find to search the
 keycode list. Adding memoize to the function wrapper caches the results
 of the search, which we don't expect to change between application
 refreshes, bringing it closer to O(1) in performance. Memory impact
 seems minimal as it's just a dictionary and results in creating less
 garbage due to the reduced searching.


jackhumbert:
Nice - thanks!

Issue #37, opened 2018-04-02
Two Click Keymapping


mechmerlin:
This request comes from notG10, one of the moderators of the Top Clack Discord.

Click and dragging can be quite strenuous. It would be great to be able to select an empty/filled spot on the keyboard representation, and then click on a keycode you want. 

To be clear, this is not to get rid of the click and drag feature, but rather as another form of keycode entry.

yanfali:
should be done. @mechmerlin can you confirm?

mechmerlin:
Yes this work! Closing.

PR #40, opened 2018-04-05
Issue #37 two click

yanfali:
- click to select a visual-keymap location and then click keycode map
     - drag keys between other keys to swap locations
     - esc to deselect a selected key
     - Animate swapping of keys on keymap
       - conveys intent of drop better
       - resize dragged key on keymap to smaller so it's obvious which is src
         - only applies to .key
       - set Z of dragged element to be on top
     - Center dragged element under cursor
     - Get originalPosition from draggable
       - dig inside draggable data and get originalPosition
     - tweak drag to 5px to make it more responsive
     - refactoring
     - add comments
     - introduce defaults object


jackhumbert:
Awesome - thanks!

PR #42, opened 2018-04-05
Use the same scaling formula everywhere

yanfali:
The larger keys use a different scaling formula than expected. Extract
this into a function and use it where needed.


jackhumbert:
Thanks!

PR #46, opened 2018-04-07
Scale Keymap dynamically

yanfali:
There is only about 775px of space in the layout for a keymap. Until we
redesign the layout to understand media queries, work within the alloted
space.

 - pre calculate the max/min of a keyboard while rendering
   - if it exceeds the available space scale the keycap size based on
     the available_space / actual_space_used.
 - introduce a config object to store copy of defaults
 - use the config object for all calculations
 - apply scale to the config object if we exceed the available space
 - extract position calculations into functions


yanfali:
screen shot 2018-04-06 at 22 27 53

Layouts that previously didn't fit, can now be rendered without spilling over

jackhumbert:
Awesome - thanks!

PR #48, opened 2018-04-08
Load readme on loading a keyboard

yanfali:
- use monakai colors for status
 - use new readme API to load keyboard markdown into status when loading
   new keyboard


jackhumbert:
Awesome - thanks!

Issue #44, opened 2018-04-06
Add dirty keymap detection


yanfali:
- centralize all access to keymap[] via some sort of API
- add hooks to detect changes
- use keymap state to decide to show alerts when switching keymaps, layouts, keyboards, importing instead of just prompting

PR #49, opened 2018-04-09
Encapsulate keymap

yanfali:
Move all keycap data structure mutation into it's own class. This is the
first step to doing dirty keymap checking before switching keymaps,
loading defaults, changing keyboard or importing a layout.

 - move responsibility for keymap data into Keymap class
 - export functions to change data inside keymap
 - modify all code to use those functions instead of direct access

Pre-requisite for issue #44


jackhumbert:
Noice. Thanks!

PR #51, opened 2018-04-09
Add two multimedia keys

corbincavolt:
Add the following multimedia keys:
- Previous \(Is Windows specific, which follows the convention set by 'Media Next'\)
- Media Stop \(Use 'Media Stop' name to differentiate from the existing application keycode 'KC\_STOP'\)


jackhumbert:
Awesome - thanks!

PR #54, opened 2018-04-14
Add keyboard bindings for selected keymap

yanfali:
Add keys mappings based on keycodes configuration.
Generates keypress.js configuration and configures library to capture
keycodes.

 - add keypress keys to keycodes list
 - generate keypress handlers from keycodes


jackhumbert:
Cool! Thanks :)

Issue #53, opened 2018-04-14
Container keycodes don't accept child keycodes


jackhumbert:
Such as `LSFT` and friends.

PR #58, opened 2018-04-18
Bug Fixes Issue #53

yanfali:
- Symptom: unable to use input text boxes
 - Cause: keypress library was intercepting all keypresses
 - Fix: narrow scope of keypress lib and turn it off for input fields in
   visual keymap

 - Symptom: unable to drag keys input key-content fields like RSFT_T
 - Cause: wrong API call sent and implementation set data to wrong part
          of keymap. Logic for identifying key_content wasn't correct
          for a contained key.
 - Fix: improve target logic to identify correct element. Fix API call
        so it is same as other instances; removed layer param. Fix API
        implementation so set content actually sets the content field.

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -109,8 +109,10 @@ $(document).ready(() => {
     urlRouteChanged();
   });
 
-  var keypressListener = new window.keypress.Listener();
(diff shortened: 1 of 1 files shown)


yanfali:
regression on key bindings

PR #59, opened 2018-04-18
Bug Fixes Issue #53 and other fixes

yanfali:
- Symptom: unable to use input text boxes
 - Cause: keypress library was intercepting all keypresses
 - Fix: make keydown handler global again and exclude input text

 - Symptom: unable to drag keys input key-content fields like RSFT_T
 - Cause: wrong API call sent and implementation set data to wrong part
          of keymap. Logic for identifying key_content wasn't correct
          for a contained key.
 - Fix: improve target logic to identify correct element. Fix API call
        so it is same as other instances; removed layer param. Fix API
        implementation so set content actually sets the content field.

 - Symptom: dragged keys float above visual keymap
 - Cause: zindex is set by draggable
 - Fix: reset z index on draggable on stop

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -112,6 +112,10 @@ $(document).ready(() => {
   var keypressListener = new window.keypress.Listener();
   keypressListener.register_many(generateKeypressCombos(keycodes));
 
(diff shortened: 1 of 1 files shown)


jackhumbert:
Awesome - thanks!

Issue #76, opened 2018-04-27
nested two-parameter keycodes fail to compile


Pawnerd:
For instance, when I put a LGui_T inside a Rctl_T keycode, the compiler comes back with an error. I think it should be impossible to put two-parameter keycodes inside another two-parameter keycode? Picture to illustrate below:

![afbeelding](https://user-images.githubusercontent.com/1347264/39339383-d43cc3e8-49c8-11e8-9f57-1c48fb4a57e9.png)

Pawnerd:
happy to help :)

yanfali:
@Pawnerd is this still an issue?

Pawnerd:
@yanfali Looks like it's fixed in the version that's running on config.qmk.fm 👍

yanfali:
@Pawnerd feel free to close the issue then. Thanks again.

PR #77, opened 2018-04-27
Issue #76 prevent nested keycodes for now

yanfali:
@jackhumbert says while there are some valid use cases for nested
container keys, we don't currently have enough UI support, a dedicated
editor, to make this easy to use or understand.

Temporarily disable the ability to nest containers within each other.


jackhumbert:
Perfect - thanks!

Issue #80, opened 2018-04-27
When importing keymap it does not set the keymap to dirty.


FrostByte-Dev:

yanfali:
@FrostByte-Dev this one should be resolved now.

Issue #74, opened 2018-04-26
layout isn't saved when we export a keyboard


yanfali:
This can lead to keymap being imported into the wrong layout and applied to the wrong keys.
steps to reproduce:

1. create a keymap
2. export it with some changes
3. change the layout
4. reimport the json

Since we didn't record the layout used to generate the map this can lead to some odd ui issues.

yanfali:
@skullydazed @jackhumbert should we add a layout field to the JSON exporter?

yanfali:
ok, I just exported a map and the layout was in there. There's obviously some bug here I need to look into.

yanfali:
OK, we suspect this is a focus issue. The input change handler never fires because the input field still has focus so it never updates the layer. @mechmerlin will verify this and we may need to use a polling loop when we focus on the field that checks for changes and updates the field if we detect one. This may cause other problems though.

yanfali:
I don't think this is an actual bug.

PR #81, opened 2018-04-27
Change layout bugs

yanfali:
- Wait for load layout to finish before changing UI.
   Uses promise returned by http GET. (Issue #74)
 - BUG change layout not called after importing a JSON
   Thanks to @FrostByte-Dev for catching this. When we loaded a JSON
   file we loaded the layout globally but never updated the route so
   this would cause the wrong layout to be selected and make the UI
   not do the right thing. (Issue #74)
 - BUG set dirty flag after importing a keymap (Issue #80)


jackhumbert:
Awesome - thanks!

Issue #92, opened 2018-05-04
warn user and prevent import of kbfirmware json files


yanfali:
When a JSON file from kbfirmware is loaded into configurator, detect the file (probably quantum field is a good choice), and pop up an alert that this is not compatible with configurator.

yanfali:
Fixed.

PR #94, opened 2018-05-05
Add tooltips and accept list for Import JSON

yanfali:
@mechmerlin recommended added tooltips to clarify what we are importing

Observed issues:

1. folks importing hex files or keymap.c files accidentally into configurator
2. folks importing kbfirmware config files into configurator

Fixes:

1. add tooltips to the import/export buttons which explain that we are expecting JSON files. Restrict the types of files we accept to application/json
2. see issue #92


jackhumbert:
Cool! Thanks :)

Issue #98, opened 2018-05-07
qmk compile failing


karni24:
trying to compile layout for hexfile but it keeps failing, have tried playing with loayout numerous times to see if that was the issue and it fails everytime

[plzwork (1).txt](https://github.com/qmk/qmk_configurator/files/1981712/plzwork.1.txt)

yanfali:
thank you! We'll take a look and see what we can figure out.

yanfali:
Hi, we may have found the culprit. Follow up question. Have you ever hand edited the JSON file? Thanks

karni24:
no, only thing ive done to the file is rename it to .txt

yanfali:
That's good. A few codes appeared in the file that aren't expected by the
API. We haven't quite figured out how they got in there. If you want to if
them. Replace the [, ], . codes in the file with KC_LBRC, KC_RBRC and
KC_STOP. It's not clear how the symbols got into the save file, so that's
definitely a bug. After I fixed those things the file compiled. Hope that
helps.

yanfali:
@karni24 good news. I've figured out how to duplicate your bug. If you configure an ANY key, that lets you type in any text into a key. This is an expert mode feature, so we don't check what you type into there. This will be saved faithfully into the keymap.json and then when you import it, because of a different feature I added for keyboard input, it imported the key incorrectly and even displayed the symbol.

To prevent this happening in the future, We're adding a check for the length of the keycode, and we'll ignore ones that are shorter than our expected length. Thanks for reporting this bug!

yanfali:
I believe this issue is now fixed. @karni24 let us know if you still have this issue, but if not please close the issue. Thanks

karni24:
thank you

PR #99, opened 2018-05-08
Issue #98 - unexpected keycode

yanfali:
While investigating the issue, we discovered a keymap that contained
invalid keycodes. The keycodes happen to parse because they matched keys
but generated invalid keymaps.

We haven't figured out how to replicate this issue, short of editing the
JSON file itself.

As a defensive measure to prevent importing bad data, add another check
for the length of the keycode. We don't support keycodes shorter than 4
characters in length. In those cases set the key to KC_NO and warn
the user on the console.

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -744,7 +744,7 @@ $(document).ready(() => {
     return key;
   }
 
-  function parseKeycode(keycode) {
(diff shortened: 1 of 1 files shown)


yanfali:
closing for now while I add a more comprehensive fix

Issue #95, opened 2018-05-07
disable download hex, source and toolbox buttons when switching keyboards.


yanfali:

yanfali:
Fixed

PR #106, opened 2018-05-11
Issue #95 Disable Toolbox, Source and Download

yanfali:
When switching keyboards, after a successful compile, we do not disable
the buttons again. This commit fixes that.

 - refactor check_status in preparation for issue #95
 - disable buttons while switching keyboards


jackhumbert:
Nice - thanks!

Issue #64, opened 2018-04-19
Exporting nameless Keymaps


mechmerlin:
Exporting a keymap without a name names it as json.txt. However this file still works when you Import it back in. 

When a nameless keymap is detected, we should default to the default "mine" name, and still create a mine.json.

yanfali:
This should be fixed now. Thanks

PR #110, opened 2018-05-12
Issue #64 Default name for empty keymap names

yanfali:
- if empty string is used for a keymap name, revert it to default on
   export to avoid it being named json.txt


jackhumbert:
Nice - thanks!

Issue #118, opened 2018-05-17
Layers that are populated with layer activating keycodes such as TO don't turn bold when clicked


mechmerlin:
Discovered while testing of PR #116.
Steps to reproduce
1. Use any keyboard/layout. 
2. Drag and drop TO and set it to a layer of your choosing. 
3. When clicking on the layer you set, a TRNS is put in the position of TO but the layer number doesn't get bolded. 

When exported and imported back in, the layer number is now bolded. 
If you don't click on that layer, it will lead to a compile error.

yanfali:
@mechmerlin this should be resolved now. Thanks

PR #119, opened 2018-05-17
Init empty layers with KC_NO

yanfali:
Fix for issue #118 

We discussed changing the default to KC_TRNS but jackhumbert felt that
for configurator this was the wrong default.

If a key with a layer change is added to a keymap it only adds a single
key to that layer. This causes problems during export and causes the
compiler to fail.

 - new empty keys are force initialized to KC_NO for a new layer
 - any action that causes a layer to be created now forces
   initialization of that layer with all keys
 - in keymap class use initLayer function for consistency
 - Change layer behavior
   - when you change layer this initializes a layer
   - layer 0 is always non-empty

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -445,6 +445,7 @@ $(document).ready(() => {
     $(e.target).addClass('active');
     layer = e.target.innerHTML;
     myKeymap.changeLayer(layer);
+    $(`.layer.${layer}`).addClass('non-empty');
     render_layout($('#layout').val());
   }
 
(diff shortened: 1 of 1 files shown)


mechmerlin:
This seems to work as advertised and I'm still able to build my keymaps as usual and compile them. This has my approval.

yanfali:
@jackhumbert I think we can merge this if it looks good to you.

jackhumbert:
Awesome - thanks folks!

PR #124, opened 2018-05-20
Move QMK toolbox link out of lower control box

yanfali:
@jackhumbert recommended moving the anchor out of the control box, and
it does look better.


jackhumbert:
Perfect - thanks!

PR #125, opened 2018-05-22
Add Atom47 default layout

Maartenwut:


jackhumbert:
This is true, but since the work is already done, I think we can go ahead and merge it for the meantime :) thanks!

Issue #114, opened 2018-05-14
Feature: Preview Info.JSON files using configurator UI


yanfali:
Add a feature which let's info.json maintainers load a info.json they are working on into the configurator UI to short cut the PR/Merge cycle review times and let contributors rapidly test their changes.

yanfali:
Basic idea is, to allow uploading an info.json and then load the same keyboard from API and extend one from the other.

e.g. $.extend(currentInfo.json, testInfo.json)

yanfali:
preview is here https://yanfali.github.io/qmk_configurator/
CTRL + SHIFT + I to load an info.json

yanfali:
Closing, as this has now been merged.

PR #134, opened 2018-05-29
Issue #114 Info.json UI preview

yanfali:
Adds the ability to preview an info.json file, which is useful for
contributors who are working on creating new info.json files. Currently
we must go through a full review and commit cycle or set up a proxy/fake
API service to preview these files.

 - add hidden keybindings CTRL + SHIFT + I to access file dialog
 - add a Fake the keybord name so it is clear in UI what is happening
   and this also prevents compilation
 - add a new onload function dedicated to previewing these files
 - modify the load layout code to handle data coming from a non API
   source
 - extract functions
   - anon onload functions extracted to named function
 - clean up direct jquery DOM lookups and use existing aliases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ secrets.tar
 .bundle/*
 vendor/*
 *.sw[po]
+node_modules
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -523,6 +523,7 @@ $(document).ready(() => {
       if (isPreviewMode) {
         $keyboard.find('option[value="'+PREVIEW_LABEL+'"]').remove();
         isPreviewMode = false;
+        setKeymapName('mine');
       }
     }
     disableOtherButtons();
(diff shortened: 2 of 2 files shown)


yanfali:
Quality Wizards @noroadsleft @mechmerlin please verify my gh-pages version for functionality. I've done some refactoring, so please retest all the things you guys usually check for, thank you. This is the version I intend to be merged upstream.

mechmerlin:
1. Each successive press of the "secret combo" results in another Preview info.json in the drop down list. This happens regardless if you select an info.json

2. No Compilation actually still outputs a message to the console `* Sending Preview info.json:info.json_preview with LAYOUT_all` implying compilation

3. After loading and testing/compiling a preview info.json. Choosing another keyboard grays out the compile button and you are unable to compile a legit keyboard. You have to reload the site to do anything.

yanfali:
1. Good catch
2. That's intentional. I can grey out compilation, there's no valid way to compile these if backend disagrees.
3. That's a side effect of 2. Removing compilation should fix this.

mechmerlin:
LGTM, I don't see any other issues.

yanfali:
Thanks QW. Good to go @jackhumbert

jackhumbert:
Cool - thanks folks!

PR #147, opened 2018-06-15
Initial Porting work to Vue 2

yanfali:
This is the first part of a planned series of PRs to port Configurator over to Vue 2.

### Why are we doing this?

As the state of the application increases, relying on state management through global variables is only going to get more difficult. We also have other goals that while possible, are made more difficult because jQuery code can become brittle during refactoring due to their being no single responsibility pattern enforcement.

These include:

 - i18n support
 - integrating more pages into the application like converter

### Why Vue?

Of the choices out there, react, angular, and ember, vue is the least disruptive and easiest to integrate into existing code. Vue strikes a gentle balance between opinionated and flexible. This has advantages when on-boarding new developers, because there are existing docs and patterns which the framework has already decided.

Other advantages:

 - good documentation
 - community developed and maintained authoritative implementations of useful components like routing and state management
 - no build process is required to begin development

### About this port

It ports responsibility for core parts of the configurator app from jquery over to vue.

This includes:
 - routing (vue-router)
 - state management (vuex)
 - introduces 2 new components
    - the status bar component used to display API status
    - the top control section of the app which contains drop downs for keyboard and layout
 - integrates the existing jQuery components into the vuex state management for interop

Testing has been performed by @mechmerlin and @noroadsleft and they feel this is now ready for deployment to production.

TODO:
  - port jquery components to vue, including:
    - status panel
    - bottom controls
  - visual keymap
  - visual keycodes

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -1,5 +1,8 @@
 $(document).ready(() => {
   const PREVIEW_LABEL = 'Preview info.json';
+  const clearKeymapTemplate = _.template(
+    'This will clear your keymap - are you sure you want to <%= action %>?'
+  );
   //  var keymap = [];
   var layer = 0;
   var job_id = '';
@@ -81,8 +84,6 @@ $(document).ready(() => {
       $fileImport.click();
     })
   );
-
-  //  $loadDefault.click(checkIsDirty(loadDefault));
 
   //Import function that takes in a JSON file reads it and loads the keyboard, layout and keymap data
   $fileImport.change(() => {
@@ -293,7 +294,7 @@ $(document).ready(() => {
     <div class="topctrl">
       <span class="topctrl-1">
       <label style="display: inline-block; width: 75px;" >Keyboard:</label>
-      <select id="keyboard" v-bind:style="width" v-bind:value="keyboard" @change="updateKeyboard">
+      <select id="keyboard" v-bind:style="width" v-model="keyboard">
         <option v-for='keeb in keyboards' :key="keeb" v-bind:value="keeb">
           {{keeb}}
         </option>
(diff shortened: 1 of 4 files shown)


yanfali:
An instance of the code is running here https://yanfali.github.io/qmk_configurator/#/1up60rgb/LAYOUT_all

AFAIK it works in both Firefox and Chrome but has not been tested in any other browser.

mechmerlin:
Looks good to me =)

assets/js/script.js
```diff
@@ -164,6 +164,11 @@ $(document).ready(() => {
     if (_.includes(keys, 'KEYMAP')) {
       return 'KEYMAP';
     }
+    // avoid keymaps ending with _kc unless we have no other choice
```
noroadsleft: IIRC, these types of keymaps don't work at all in the Configurator. I think we should exclude them completely unless we're planning to support them in future, which I don't think we are.

Also, I know a couple keyboards have these types of keymaps, but the macro *contains* `_kc_` rather than ending in `_kc`. I know the Contra is one of those.

assets/js/script.js
```diff
@@ -165,13 +175,17 @@ $(document).ready(() => {
       return 'KEYMAP';
     }
     // avoid keymaps ending with _kc unless we have no other choice
```
noroadsleft: IIRC, these types of keymaps don't work at all in the Configurator. I think we should exclude them completely unless we're planning to support them in future, which I don't think we are.

Also, I know a couple keyboards have these types of keymaps, but the macro *contains* `_kc_` rather than ending in `_kc`. I know the Contra is one of those.

yanfali: Thanks @noroadsleft. I'll file an issue about this soon.

Issue #153, opened 2018-06-17
look at why planck doesn't compile


yanfali:

noroadsleft:
Culprit is the `EXT_PLV` keycode. https://github.com/qmk/qmk_configurator/blob/ee789f0a895d62df4384ae15ee673a56d0c4e751/keymaps/planck_default.json#L11

Compiles without errors if you delete the keycode.

----

Defined in as a macro in qmk_firmware.

- https://github.com/qmk/qmk_firmware/blob/44660e77a4e4c3f6ab180df1f8135ba0691dc012/keyboards/planck/keymaps/default/keymap.c#L38
- https://github.com/qmk/qmk_firmware/blob/44660e77a4e4c3f6ab180df1f8135ba0691dc012/keyboards/planck/keymaps/default/keymap.c#L152
- https://github.com/qmk/qmk_firmware/blob/44660e77a4e4c3f6ab180df1f8135ba0691dc012/keyboards/planck/keymaps/default/keymap.c#L238

yanfali:
@jackhumbert what do you want to do? If we leave this out the keymap will compile in configurator.

yanfali:
@noroadsleft you want to open a PR with the code removed and let's get @jackhumbert to review it?

noroadsleft:
I'm willing to do so, but then the default in Configurator won't match the default in qmk_firmware, which I'm not keen on.

I have no idea how the Plover support works, so can't help there.

yanfali:
Yeah, I think it's unavoidable though for the time being unless we want a
non working config.

yanfali:
Awesome work @noroadsleft

noroadsleft:
Thank you, @yanfali.

PR #160, opened 2018-06-18
Removed EXT_PLV keycode from Planck default keymap

noroadsleft:
Re: #153 

The `EXT_PLV` keycode on Layer 5 causes the Planck default keymap to fail to compile in the Configurator, because it's a macro assignment, rather than a QMK-native keycode. (Indeed, it appears based on my testing, that the `ANY()` functionality only works if the keycodes included are QMK-native.)

Tagging @yanfali for context and @jackhumbert for review.


yanfali:
Thank you @jackhumbert and @noroadsleft this one has been bugging me for a while.

PR #161, opened 2018-06-19
Add defaultish keymaps for Preonic and Atomic

noroadsleft:
Matched the current default keymaps from the qmk_firmware repo as closely as I could.

Could add Subatomic, if that default is (close to) finalized.

Tagging @jackhumbert because they're his boards.


jackhumbert:
Cool! Thanks :D

PR #174, opened 2018-07-11
Add default ANSI hs60 keymap

mechmerlin:
That's right, you read it correct. ANSI. =)


yanfali:
@yiancar you ok with this?

yiancar:
whatever the magician says

PR #176, opened 2018-07-17
Display a random potato factoid

yanfali:
At the suggestion of @mechmerlin we're adding a random potato fact to
the tagline of the configurator page. Mostly to show how amazing
potatoes are and to show how much QMK loves the humble potato.


jackhumbert:
Perfect - thanks :D

PR #191, opened 2018-09-03
Feature: filter keyboard list using a prefix

yanfali:
By appending a `?filter=<prefix>` parameter a user can restrict the
restrict which keyboards are being served from the QMK configurator
interface to a subset that makes sense for them.

Examples
========

  1. First Load: `config.fm.qmk/#/?filter=planck` loads only planck
     variants.
  1. Dynamically filter keyboard list. Append `?filter=clue` to URL to
     show only clueboards.
  1. Alphabetical filter. Append `?filter=d` to URL to only see
     keyboards starting with d.
  1. Exact filter. Append `?filter=hs60` will only display a single
     keyboard in the drop down.

Other
=====

 - Update index.html title attribute

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -158,6 +158,7 @@ $(document).ready(() => {
       state: {
         keyboard: '',
         keyboards: [],
+        _keyboards: [],
         layout: '',
         layouts: {},
         keymapName: '',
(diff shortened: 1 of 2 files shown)


jackhumbert:
Awesome :) thanks!

PR #193, opened 2018-09-18
Update code to use new URLs embedded in response

yanfali:
- instead of the client calculating the URL, let the server tell us
 - expects an array of results. In the future we may extend this to
 allow downloading different formats for variants like the tada68 which
 can be flashed with DFU to accept a hex file

@noroadsleft @mechmerlin please apply your QA Wizarding hats to this on https://yanfali.github.io/qmk_configurator/ @skullydazed and I have lightly tested this with .hex and .bin targets. Tada68 should now be fixed for the default target.

We are also one step closer to getting ARM support and potentially allowing multiple download targets (e.g. hex or bin)

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -166,8 +166,8 @@ $(document).ready(() => {
         isPreview: false,
         jobID: '',
         enableDownloads: false,
-        firmwareFile: '',
-        firmwareStream: '',
+        firmwareBinaryURL: [],
+        firmwareSourceURL: [],
         filter: ''
       },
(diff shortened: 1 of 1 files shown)


mechmerlin:
Everything seems to work like normal. The downloaded bin file is no longer "null", but I'd like to see someone actually flash it on a Tada.

yanfali:
@jackhumbert @skullydazed should be ready to merge

skullydazed:
Thanks!

PR #198, opened 2018-09-30
Issue #175 Add keyboard folder exclusion to dropdown

yanfali:
@mechmerlin @noroadsleft time for you guys to do some QA Wizarding

PR #199, opened 2018-10-01
Added bigseries and clueboard/66 to keyboard exclusion list

noroadsleft:
Noticed a couple "keyboards" that seem to have been missed in the review stage for #198.

`bigseries` was a parent directory for the Big Series line of PCBs by Woodkeys, and `clueboard/66` was a parent directory for revs. 1-3 of the Clueboard 66%.

@yanfali @mechmerlin

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -2365,8 +2365,10 @@ $(document).ready(() => {
   function getExclusionList() {
     return [
       'atom47',
+      'bigseries',
       'chibios_test',
       'christmas_tree',
+      'clueboard/66',
       'crkbd',
       'deltasplit75',
       'duck/eagle_viper',


noroadsleft:
Good spot, @mechmerlin.

skullydazed:
Why are we excluding `clueboard/66`?

yanfali:
I think because on the configurator we'd prefer people to know what rev they are hexing. There's been issues with people take a different rev hex and having stuff not quite working in the past. Like rev3 having more RGB than previous versions. At least that's the rationale I've heard mechmerlin espouse.

skullydazed:
So I don't disagree with that logic. But I don't want there to be a difference between CLI and API this way, at least for my own boards. If we want to eliminate this ambiguity the place to do that is in qmk_firmware, not by making the API different from the CLI.

yanfali:
I'm fine with that. If you want to revert this and fix it in qmk_firmware that sounds like a good plan. We still have to deal with top level folders that don't build in the mean time. What would you like to do specifically for the clueboard/66 ?

PR #204, opened 2018-10-19
Add default-ish keymap for Massdrop CTRL

noroadsleft:
Requested in qmk/qmk_firmware#4175 by @jamesalmeida

----

Adapted from https://github.com/qmk/qmk_firmware/blob/78b48371aac3bc4ba468c9f7efcb742bafb81965/keyboards/massdrop/ctrl/keymaps/default/keymap.c

Caveats:

Many of the keycodes used in the default keymap are actually macro keys specific to the CTRL, rather than QMK-native keycodes. I have removed these keycodes and replaced them with `KC_NO` (no operation). All of these keycodes are on the Function layer.

Configurator's API doesn't support macros yet, so this is as close as can be had, but it should still be a good starting point for anyone with a CTRL.


yanfali:
Verified by @patrickmt. Thanks very much the both of you.

PR #205, opened 2018-10-19
Restructure default keymaps for Planck

noroadsleft:
# Restructure default keymaps for Planck

Due partially to changes in the Configurator's API, only the Planck rev6 has a default keymap JSON according to the Configurator.

`planck_default.json` has been renamed to `planck_light_default.json`, and had its JSON contents updated to `keyboard: planck/light`. This file was then copied and updated to support Planck in revs. 3, 4 and 5.

All JSON files also now reference the `LAYOUT_ortho_4x12` matrix, rather than the deprecated `KEYMAP`.

# Convert to Unix line endings

Per .editorconfig.


yanfali:
Awesome. Thank you.

PR #206, opened 2018-10-19
Restructure default keymaps for Preonic

noroadsleft:
Similar to #205. API changes have broken the functionality provided by these files.

----

# Rename/copy Preonic files

Updated to use the filenames now expected by the API.

# Update keyboard names in JSON data

# Convert to Unix line endings

Per .editorconfig.


yanfali:
Thank you

PR #208, opened 2018-10-22
New 1up60hse default keymap

mechmerlin:
1. My change to include all 1up keyboards in its own 1upkeyboards directly broke this default keymap.
2. skiwithpete has requested we update the default keymap to a new one he made.


yanfali:
Thank you.

PR #209, opened 2018-10-26
Remove Repetitive Button Labels

### LINKED ISSUE #194: Better Button Naming
### Examples
For example, in the RGB section, we have buttons called RGB something something. 
- We can do Mode+ and Mode-. instead of RGB Mode +. After all we do have just Hue+ and Hue-
- Rather than RGB Mode B, we could just call it Breathe. 
- We could call RGB Mode R, just Rainbow. 

### Issues
There might be overlaps, if we ever do implement backlight breathing, then we would have two buttons labeled breathing. Hopefully people would be able to determine which is which because one is in the Backlight Section and one is in the RGB section. The same goes for Toggle.

MarceloAlves:
Renamed buttons so that they're not repetitive.

Also changed `.keycode` `width` to `min-width: 31px;` so that longer named buttons aren't breaking out of the square. Just thought I'd throw that in there. You're choice and whether to keep it or not!

From
screen shot 2018-10-25 at 5 38 23 pm

To
screen shot 2018-10-25 at 5 38 56 pm

Closes #194


yanfali:
Hi, thanks for working on this. We're about to do a redesign so I'm not sure how valuable this is moving it forward.

It also breaks the uniform sizes of the key code list, so it feels like the wrong direction. We've prototyped a tabbed UI in the past, but I think grouping in tabs would help much more than moving things around.

yanfali:
I also think this might be more confusing in the keymap as it'll display the name

MarceloAlves:
@yanfali Okay no worries! I’ll go ahead and close this then. I noticed the rewrite in Vue, I’ll keep an eye out for places I can contribute. I’ve been wanting to mess around with Vue 👍🏽

PR #210, opened 2018-10-26
Add default layout for bpiphany/unloved_bastard LAYOUT_tkl_ansi

waynr:


waynr:
@yanfali thanks for catching that! So I just kind of blindly tapped in the key codes from my keyboard in the configurator UI and exported them. Currently I am running qmk firmware in this keyboard with the following definition for my default layer:
```
[0] = LAYOUT(\
                 KC_ESC, KC_F1,  KC_F2,  KC_F3,  KC_F4,  KC_F5,  KC_F6,  KC_F7,  KC_F8,  KC_F9,  KC_F10, KC_F11,  KC_F12,             KC_PSCR, KC_SLCK, RESET, \
                 KC_GRV, KC_1,   KC_2,   KC_3,   KC_4,   KC_5,   KC_6,   KC_7,   KC_8,   KC_9,   KC_0,   KC_MINS, KC_EQL,  KC_BSPC,   KC_INS,  KC_HOME, KC_PGUP, \
                 KC_TAB, KC_Q,   KC_W,   KC_E,   KC_R,   KC_T,   KC_Y,   KC_U,   KC_I,   KC_O,   KC_P,   KC_LBRC, KC_RBRC, KC_BSLS,   KC_DEL,  KC_END,  KC_PGDN, \
                 LT(1, KC_CAPS),KC_A,   KC_S,   KC_D,   KC_F,   KC_G,   KC_H,   KC_J,   KC_K,   KC_L,   KC_SCLN,LT(1, KC_QUOT),          KC_ENT, \
                 KC_LSFT,KC_NUBS,KC_Z,   KC_X,   KC_C,   KC_V,   KC_B,   KC_N,   KC_M,   KC_COMM,KC_DOT, KC_SLSH,          KC_RSFT,            KC_UP, \
                 KC_LCTL,KC_LGUI,KC_LALT,                KC_SPC,                                 KC_RALT,TG(1), KC_RCTL,  KC_APP,   KC_LEFT, KC_DOWN, KC_RGHT  )
```

So aside from the obvious fact that I have the "Right OS" key set to toggle layer 1, I also have the keys labeled as "CTRL" and "FN" swapped. Even so, according to the definition above it should still be "Right CTRL" aka `KC_RCTL`, but it seems to be detected as "Left CTRL" aka `KC_LCTL`  when I hit that key in the configurator GUI.

I'll fix this up and submit a more correct version shortly, but this seems to indicate that either there is a bug somewhere in the unloved_bastard firmware stack (unlikely) or I used the wrong keyboard definition when customizing it originally. Taking a closer look at the two available keymaps in qmk_firmware master branch for this firmware, there is `default` and `default_ansi`. I am using `default`, and it looks like I should be using `default_ansi`. I'll try recompiling and dig in further if that doesn't work.

Issue #211, opened 2018-10-31
Bug: info.json upload for preview only works once


noroadsleft:
The Configurator feature that allows a user to upload a file to preview how it will render when merged (assuming no JSON-to-keyboard.h mismatches) by pressing Ctrl+Shift+I. It has had a bug of sorts for a while where it only worked once, and you had to refresh the page to use it again. I recently stumbled upon why that is.

When the user presses Ctrl+Shift+I, the HTML select element for `Keyboard` gains a `Preview info.json` option element as soon as the shortcut is used. If the user uploads a file, the preview loads as expected, but if the user cancels the upload, the element remains, but the select element remains on whichever keyboard was already selected.

Changing to the Preview option manually and then changing off of it removes the preview element, allowing the user to use the shortcut again.

Ideally, the preview option shouldn't be added unless the user actually uploads a file.

Additionally, the user should be allowed to upload multiple times without having to refresh the page, but that may be a separate issue.

yanfali:
This should be fixed now.

noroadsleft:
Confirmed as fixed.

PR #212, opened 2018-10-31
Upload info.json preview state bug

yanfali:
Separate the act of requesting a preview, from the state of being in a
preview. Input file doesn't send a cancel callback if the user opens a
file dialog but then cancels it. Add a new state where we preview is
requested which is only used to open the dialog and it is reset
immediately after.

Thanks to @noroadsleft for figuring this tricky state issue out.

Fix for issue #211 

Calling @noroadsleft @mechmerlin for some quality wizarding


noroadsleft:
Tested and working for me on @yanfali's fork.

Issue #214, opened 2018-11-09
Can't write in "Any" blocks


ThibaultMondanel:
These blocs turn blank as soon as you change the layer level
![image](https://user-images.githubusercontent.com/8020057/48245825-d7475a80-e3ba-11e8-9394-658ee4c7e10f.png)

ThibaultMondanel:
I tried these codes:
KC_SYSTEM_SLEEP ; KC_SLEP ; KC_BTN1 ; KC_BTN2 ; KC_BTN3

yanfali:
the technical reason any is leaving the codes blank is it's checking them against a built table of codes and ignoring ones it doesn't know about. Not the best user experience I know sorry.

yanfali:
OK, SLEP is in the UI now. I'm going to close this bug as resolved. Thanks for reporting it @ThibaultMondanel

PR #216, opened 2018-11-09
Add the Sleep media code

yanfali:
This one was requested by a user on #help channel and it seems harmless to add it.

@mechmerlin what do you think?


yanfali:
Helps with issue #214

PR #226, opened 2018-12-14
Updated Keyboard Exclusion List: 2018-12-12 patch 1

noroadsleft:
Mistyped the exclusion rule for converter/sun_usb.


yanfali:
Thanks

PR #237, opened 2019-01-22
No support for #define in configurator API

yanfali:
_FL -> 1

@TerryMathews


TerryMathews:
Thank you. I was made aware of this around noon. I was going to submit a PR
when I got home from work.

Thanks for the quick fix!

PR #242, opened 2019-02-01
add MacOS JP key and rename JP Kana key

takashiski:
There is no alternate key, so I add MacOS JP keys, Kana and Eisu.

And I change JIS Kana name to  "カタカナ\nひらがな(Katakana Hiragana)" from "かな(Kana)" because many keyboard has it as Katakana Hiragana. 

Reason of inserting "\n" is Japanese characters is not be line break by space.


takashiski:
Awesome! Beta has developed on your repo?
https://yanfali.github.io/qmk_configurator/

yanfali:
Yes. One more week and we go production.

yanfali:
@takashiski we're working on a JP translation of the configurator. If you're interested we could use some help https://github.com/yanfali/qmk_configurator/tree/yanfali-i18n

PR #244, opened 2019-02-09
Initial defaultish layout for TKC M0LLY

TerryMathews:


yanfali:
Thanks for this. We're about to push out the new version of configurator, so I'll hold off merging this for a day or two if that's ok with you.

TerryMathews:
I don't want it to be an inconvenience, but the other prototype holder
needs config.qmk.fm. He's not savvy enough to compile from source.

I'll gladly take ownership and change the keymap file if the new version
breaks it.

yanfali:
OK, understood. It's not going to work until your get the PR from qmk_firmware merged though, so it may take a day or two anyway.

PR #245, opened 2019-02-09
Fixed default keymap inconsistency

mattdibi:
Hi,
recently I was brought to the attention of the missing `?` key in the default keymap of the Redox keyboard. Here's the fix which reflects more the default keymap in the QMK firmware repository.


yanfali:
Awesome. Thanks

Issue #253, opened 2019-02-17
API behavior has changed on error


yanfali:
![image](https://user-images.githubusercontent.com/2275667/52910778-3c7bd900-3251-11e9-8226-301391c58095.png)

https://cdn.discordapp.com/attachments/440875445911683084/546563933616472083/basic_dz60.json

Change the UI code to check is_failed before enabling download firmware button. The keymap and source downloads should be enabled though for diagnostic reasons. Just check that the firmware url is actually empty and don't enable the download firmware.

yanfali:
[keymap-dz60-basic_dz60.zip](https://github.com/qmk/qmk_configurator/files/2872789/keymap-dz60-basic_dz60.zip)
Example of JSON file that triggers an error on API

yanfali:
should be resolved now. Closing

PR #254, opened 2019-02-18
BUG: API behavior refined to support src

yanfali:
Even when a keymap doesn't compile we now support downloading the
keymap.c and source.

 - use vuex helpers
 - add computed properties for each URL property
 - allow download firmware to remain disabled


yanfali:
Corrects #253

PR #256, opened 2019-02-21
Add LT Support to Configurator

yanfali:
Rather than a complicated dual input UI, I've decided to go for something a tad simpler. Pre-defined LT codes that have the layer value preset. We only support 16 layers in the UI so it's simpler conceptually to map it more like LCTL(KC) and friends where it becomes LT(1,KC) and we only allow the user to change KC but the layer is already incorporated.

Tested:
  - UI
  - import existing keymaps
  - exporting new keymaps
  - compilation

### DIFF
--- a/public/style.css
+++ b/public/style.css
@@ -983,12 +983,14 @@ button {
 }
 
 .keycode-container,
-.keycode-layer {
+.keycode-layer,
+.keycode-layer-container {
   font-size: 10px;
   display: block;
 }
 
 .keycode-container:after,
+.keycode-layer-container:after,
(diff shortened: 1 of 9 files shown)


yanfali:
@mechmerlin @noroadsleft can you guys please confirm that you've done some light testing of the new feature and that you think it's ready for prime time.  Thank you both.

yanfali:
Screen shot of new UI element
screen shot 2019-02-21 at 08 56 28

Screenshot of how it looks on the visual keymap
screen shot 2019-02-21 at 08 55 30

PR #257, opened 2019-02-22
Add application launch and browser keycodes

richarddewit:


yanfali:
@richarddewit awesome. Couple of nits. Can you fix the spacing?

screen shot 2019-02-22 at 13 00 23

1. Nudge Exec over a bit and align it with Browser Search I think. It just looks a bit weird with that spacing.

2. There are linting errors in the code. Please fix those. We use prettier for auto formatting, so you may have to set up your environment, but there maybe a way to do it from the command line.

Other than that it looks great, and thank you for this.

PR #258, opened 2019-02-24
Add defaultish keymap for Massdrop Alt

AyyKamp:
[Based on the original keymap.c](https://github.com/qmk/qmk_firmware/blob/4d9b11af14fd83c1e6b556165d41ce005abb4a84/keyboards/massdrop/alt/keymaps/default/keymap.c) except for the Massdrop specific keycodes as they are afaik not supported by the API. I have replaced them with `KC_NO` to be consistent with the Massdrop Ctrl defaultish keymap.

By the way, is it planned to add support for these keyboard specific keycodes?


yanfali:
Thanks, that's awesome. The onus is on massdrop to get their codes added, qmk has to support 200+ keyboards so we'll have to figure it out. It is possible we can support special codes for certain boards based on the info.json. we're talking about it for steno keyboards for example.

AyyKamp:
I see, thanks for the answer :)

PR #276, opened 2019-03-11
Added a Steno key tab

germ:
The only issue I see with this PR is that it requires the user to already have QK_STENO_GEMINI and whatnot done up. I guess that also goes for any of the Command keys and other feature dependent keycodes.


yanfali:
Thanks for this. As discussed on discord I think we should probably hide this behind a white list as it only really applies to a small number of boards and requires it to be turned on. I'll send you a PR against this PR that adds the necessary infrastructure to selectively enable the feature.

yanfali:
https://github.com/germ/qmk_configurator/pull/1 is more of how I think this feature should work for now until we improve info.json

PR #277, opened 2019-03-12
Tweak the CSS grid for ControllerTop.vue

noroadsleft:
`cannonkeys/satisfaction75/prototype` was added to QMK Firmware, and now the Configurator's UI rendering in Firefox is wonky because the Keyboard select element is so wide.

Added 40px of width to the left and middle columns of the grid.

ui_tweak_before
ui_tweak_after

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -320,7 +320,7 @@ export default {
 <style>
 .topctrl {
   display: grid;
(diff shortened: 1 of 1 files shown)


yanfali:
Not sure about. This try setting overflow to hidden and just truncating the text.

yanfali:
Thanks for doing this, closing in favor of #278

PR #278, opened 2019-03-12
Fix max width of keyboard element

yanfali:
@noroadsleft I think this is probably the way we should go for now.

- also cleaned up CSS technical debt using Violet Pena's recommendations
  @see https://vgpena.github.io/using-css-grid-the-right-way/

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="controller-top">
     <div class="topctrl">
-      <div class="topctrl-1-1">
+      <div class="topctrl-keyboards">
         <label class="drop-label">Keyboard:</label>
         <select id="keyboard" v-bind:style="width" v-model="keyboard">
           <option v-for="keeb in keyboards" :key="keeb" v-bind:value="keeb">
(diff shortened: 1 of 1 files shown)


src/components/ControllerTop.vue
```diff
@@ -309,7 +309,7 @@ export default {
   data: () => {
     return {
       keymapName: '',
-      width: 0
+      width: 'max-width: 18rem'
```
noroadsleft: This is the only part of this change I don't understand how it works.

\- @noroadsleft 
*newb who doesn't understand the dark magic that is Vue*

yanfali: yeah, it's basically a component value that is templated into the page. We can move it into the css. I think the original intention was the size of the element was dynamically adjusted based on the width of the text. I got rid of that feature but didn't remove the templating.

yanfali: I'm going to use the CSS you proposed

noroadsleft: We can revisit this method later on when I understand wtf is going on here.

yanfali: oh yeah for sure. I actually used your css and removed the code.

yanfali:
Verbal agreement from @noroadsleft

Issue #282, opened 2019-03-16
warning for layer keys


yanfali:
Warn user in keycode UI that layer keys overwrite the layer ppl position above to avoid locked layers.

PR #283, opened 2019-03-16
Provide a warning for Layer Keys

yanfali:
- enhance space key to support optional title and icon and hover
 - provide a warning for the layer keys that using them will overwrite
   the key above to KC_TRNS to prevent layer lock
 - add fa-exclamation-triangle icon

@fauxpark @noroadsleft @mechmerlin

### DIFF
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -37,7 +37,7 @@ export default [
     iconClass: 'warning',
     width: 'label',
     title:
-      'Note: Layer keys override the position on the target layer to TRNS to avoid layer lock'
+      'Note: Layer keys overwrite the key on the target layer to KC_TRNS to avoid layer lock'
   },
 
   {
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -37,7 +37,7 @@ export default [
     iconClass: 'warning',
     width: 'label',
     title:
-      'Note: Layer keys override the switch position above to TRNS to avoid layer lock'
+      'Note: Layer keys override the position on the target layer to TRNS to avoid layer lock'
   },
 
   {
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -31,7 +31,14 @@ export default [
     title: 'Manually enter any QMK keycode'
   },
 
-  { label: 'Layer and Layer Tap functions', width: 'label' },
+  {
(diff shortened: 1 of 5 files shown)


yanfali:
Warning Icon

Warning Text

yanfali:
Issue #282

noroadsleft:
I'm thinking saying "on the target layer" instead of "above" is better. "Above" makes it sound like the KC_TRNS is forced onto the next layer, but the next layer might not be the target layer (user could be doing MO(3) from Layer 0).

Maybe in a future iteration, we could notify the user which keycode was removed, if the forced KC_TRNS overwrites a key.

src/store/modules/keycodes/quantum.js
```diff
@@ -31,7 +31,14 @@ export default [
...
   },
 
-  { label: 'Layer and Layer Tap functions', width: 'label' },
+  {
+    label: 'Layer and Layer Tap functions',
+    icon: 'exclamation-triangle',
+    iconClass: 'warning',
+    width: 'label',
+    title:
+      'Note: Layer keys override the position on the target layer to TRNS to avoid layer lock'
```
fauxpark:
```suggestion
      'Note: Layer keys overwrite the key at the same position on the target layer to KC_TRNS to avoid layer lock'
```

"Override" implies the key will return to its original assignment if the layer key is removed.

yanfali: Sorry that makes no sense unless we allow undo.

fauxpark: Uh, it *does* make sense *because* undo is not supported (yet).

yanfali: I do like the overwrite better though. I will need to test to see if it's too verbose for the screen. It's quite long already and on a contricted width it may be too wordy

yanfali: Agree to disagree. But thank you for the input.

fauxpark: I'm confused as to what we're disagreeing on... you say you like "overwrite" better, which is exactly what the suggestion does.

fauxpark: For clarity, my comment below the suggestion is about the current behaviour, where if you remove a MO(), the transparent key above it is not also removed or reverted back to what it was before the placement of the MO() turned it into KC_TRNS. I call that "overwriting", not "overriding".

yanfali: Thanks very much.

PR #293, opened 2019-03-22
Restyle Keymap Name and Layout form fields

noroadsleft:
### Restyle Keymap Name and Layout form fields (49ecb7c)

Restyle Keymap Name and Layout form fields to match searchable Keyboard field.


yanfali:
Thanks

PR #303, opened 2019-04-02
Capture KC_APP properly

fauxpark:
In Chrome and Safari on macOS, `ev.keyCode` for the left OS and context menu key is the same, and Keypress.js favours the former, so we need a second opinion from `ev.key`.

I *think* this is the right place to put this...

Side note: On Safari, the context menu key is actually not even recognised and [results in `KC_BSLS`](https://github.com/dmauro/Keypress/blob/df626c49f3208386a735d0bb1bd38b7aafbc1b02/keypress.coffee#L806):

```coffeescript
_keycode_dictionary = 
    0   : "\\"          # Firefox reports this keyCode when shift is held
    8   : "backspace"
    9   : "tab"
```

The Configurator should probably throw away any key events where `e.key === 'Unidentified'`.

### DIFF
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -45,6 +45,10 @@ function generateKeypressCombos(_keycodes) {
   return combos;
 }
 
+const keyLUT = {
+  'ContextMenu': 'KC_APP'
+};
+
 const mods = {
   KC_LSFT: 'KC_RSFT',
   KC_LCTL: 'KC_RCTL',
@@ -59,21 +63,23 @@ function generateKeypressHandler(keycode) {
     keys: keycode.keys,
     on_keydown: ev => {
       let _meta = meta;
-      // detect left and right mods
-      switch (meta.code) {
-        case 'KC_LSFT':
-        case 'KC_LGUI':
-        case 'KC_LALT':
-        case 'KC_LCTL':
-          // Keypress reports this keycode as Left GUI because they (sometimes) have the same event keyCode
-          if (ev.key === 'ContextMenu') {
-            _meta = store.getters['keycodes/lookupKeycode']('KC_APP');
+
+      // handle special cases eg. ContextMenu
+      const special = keyLUT(ev.key);
+      if (!isUndefined(special)) {
+        _meta = store.getters['keycodes/lookupKeycode'](special);
+      } else {
+        // detect left and right mods
+        switch (meta.code) {
+          case 'KC_LSFT':
+          case 'KC_LGUI':
+          case 'KC_LALT':
+          case 'KC_LCTL':
+            if (ev.location === ev.DOM_KEY_LOCATION_RIGHT) {
+              _meta = store.getters['keycodes/lookupKeycode'](mods[meta.code]);
+            }
             break;
-          }
-          if (ev.location === ev.DOM_KEY_LOCATION_RIGHT) {
-            _meta = store.getters['keycodes/lookupKeycode'](mods[meta.code]);
-          }
-          break;
+        }
       }
 
       store.commit('keymap/setKeycode', { _code: _meta.code });
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -46,7 +46,7 @@ function generateKeypressCombos(_keycodes) {
 }
 
 const keyLUT = {
-  'ContextMenu': 'KC_APP'
+  ContextMenu: 'KC_APP'
 };
 
 const mods = {
@@ -65,7 +65,7 @@ function generateKeypressHandler(keycode) {
       let _meta = meta;
 
       // handle special cases eg. ContextMenu
-      const special = keyLUT(ev.key);
+      const special = keyLUT[ev.key];
       if (!isUndefined(special)) {
         _meta = store.getters['keycodes/lookupKeycode'](special);
       } else {
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -65,6 +65,11 @@ function generateKeypressHandler(keycode) {
         case 'KC_LGUI':
         case 'KC_LALT':
         case 'KC_LCTL':
+          // Keypress reports this keycode as Left GUI because they (sometimes) have the same event keyCode
+          if (ev.key === 'ContextMenu') {
+            _meta = store.getters['keycodes/lookupKeycode']('KC_APP');
+            break;
+          }
           if (ev.location === ev.DOM_KEY_LOCATION_RIGHT) {
             _meta = store.getters['keycodes/lookupKeycode'](mods[meta.code]);
           }


yanfali:
Also, we don't official support Safari, we only make reall test on Firefox and Chrome. Safari is less than 9% of the traffic we get on configurator.

yanfali:
I think we just need to break out ev.key into it's own branch. Rather than overloading the mod detection. This is basically it's own path. Right now, this patch depends on a bug in the way the library above it is detecting things and overloading mod detection.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

My preference would be add a lookup using a match object:

```javascript
keyLUT = { 'ContextMenu': 'KC_APP' }

/* then later we do */
const special = keyLUT(ev.key);
if (!isUndefined(special)) {
  _meta = store.getters['keycodes/lookupKeycode'](special);
}
```

That's a more generic mechanism and we can use it for things like AltGr detection and what not. Do you think that would solve your problem?

fauxpark:
> Also, we don't official support Safari, we only really test on Firefox and Chrome. Safari is less than 9% of the traffic we get on configurator.

I figured, but the main point there was that unrecognised keys (regardless of browser, should have clarified) will erroneously insert a `KC_BSLS` instead of doing nothing.

> Right now, this patch depends on a bug in the way the library above it is detecting things and overloading mod detection.

Technically, Keypress.js is correct, because the browsers I mentioned above do use that `keyCode` value for the GUI key. It just assumes that the context menu key will never be pressed under those conditions since Mac keyboards don't have that key anyway.

> My preference would be add a lookup using a match object:

Where would I add this?

yanfali:
I would probably try and add it before the mod detection and just return if it's a match. Put the lookup table outside in a const block, probably near the code that uses it. Then we just need to test it. A lot of people just redefine that key to be a layer key, so I'm not entirely sure how many people even use this feature. Maybe the API can start recording a frequency map of keycodes across compiles.

yanfali:
This shouldn't ever match anything else right @fauxpark ?

fauxpark:
Not sure what you mean there. `KC_APP` appears to == `ContextMenu`, there's nothing else that key name is used for.

How is this:
```diff
diff --git a/src/jquery.js b/src/jquery.js
index c4c5841..6087330 100644
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -45,6 +45,10 @@ function generateKeypressCombos(_keycodes) {
   return combos;
 }
 
+const keyLUT = {
+  'ContextMenu': 'KC_APP'
+};
+
 const mods = {
   KC_LSFT: 'KC_RSFT',
   KC_LCTL: 'KC_RCTL',
@@ -59,22 +63,24 @@ function generateKeypressHandler(keycode) {
     keys: keycode.keys,
     on_keydown: ev => {
       let _meta = meta;
+
+      // handle special cases eg. ContextMenu
+      const special = keyLUT(ev.key);
+      if (!isUndefined(special)) {
+        _meta = store.getters['keycodes/lookupKeycode'](special);
+      } else {
         // detect left and right mods
         switch (meta.code) {
           case 'KC_LSFT':
           case 'KC_LGUI':
           case 'KC_LALT':
           case 'KC_LCTL':
-          // Keypress reports this keycode as Left GUI because they (sometimes) have the same event keyCode
-          if (ev.key === 'ContextMenu') {
-            _meta = store.getters['keycodes/lookupKeycode']('KC_APP');
-            break;
-          }
             if (ev.location === ev.DOM_KEY_LOCATION_RIGHT) {
               _meta = store.getters['keycodes/lookupKeycode'](mods[meta.code]);
             }
             break;
         }
+      }
 
       store.commit('keymap/setKeycode', { _code: _meta.code });
     }
```

We would also need to add another if statement above that to `return` when `ev.key === 'Unidentified'`, the lookup table won't work for that.

yanfali:
That looks pretty good. Does it work for you? I'm not too worried about Unidentified, as it wouldn't match and just fall through to the normal detection and binding. I'll add KC_APP to of my layer keys to test it.

yanfali:
I'm your branch, push and I can test it for you too.

fauxpark:
I've not tested this, looking at getting nvm set up now. There are apparently some caveats installing via homebrew.

yanfali:
couple of little mistakes
Couple of tiny issues

keyLUT[] instead of keyLUT(), that's my typo, it's an object lookup.
You can also remove the quotes around the key, it's a object, doesn't need to be in JSON proper form for ContextMenu key.

Let me test those changes locally.

yanfali:
Works great with those small changes. I tested this in FF66 and Cr73. So the logic is basically fall through. Each key binding already has a associated keycode, these branches are basically exceptions. If the LUT doesn't match it hits mod detection, if the mods don't match it uses the default value. Unidentified won't match anything so it won't cause any change.

fauxpark:
> keyLUT[] instead of keyLUT(), that's my typo, it's an object lookup.

Got it, thought that might be some newfangled Javascript thing or something.

> Unidentified won't match anything so it won't cause any change.

Unidentified matches `KC_BSLS` because they both have `keyCode` 0.

yanfali:
@fauxpark but that's only on Safari right? I don't know if it's even worth fixing.

fauxpark:
Not just Safari, it just happens that it refuses to recognise the context menu key and so generates an `Unidentified` KeyboardEvent. There are probably other virtual keycodes that can't be mapped.

yanfali:
@fauxpark it's fine, every keybinding has a default already, so it'll do what the default is. I can't fix every browser in the world. Unidentified just gets ignored. We've never checked it before, so what's different now? We're whitelisting a single code.

fauxpark:
It *doesn't* get ignored though, that's the thing. It inserts a `KC_BSLS` because they have the same `keyCode`.

yanfali:
But it's correct for that browser. So tough luck, get a better browser.

fauxpark:
What do you mean it's correct for that browser? Which browser? I'm saying that a KeyboardEvent with `ev.key === 'Unidentified'` should do absolutely nothing in the Configurator, because there's nothing useful we can do with that event. Instead, *regardless of the browser or OS*, that KeyboardEvent would insert a `KC_BSLS`.

yanfali:
I'm arguing we never looked at it in the first place, so it doesn't make any difference. It's not any different from what happens now. It's not a bug. Your browser is broken. I'm OSX. I tested your patch on Cr and Firefox it works.

fauxpark:
Perhaps I should clarify that I don't use Safari, I just tested it out of curiosity. An `Unidentified` KeyboardEvent is not Safari-specific. It will occur whenever the browser generates a KeyboardEvent but can't map the virtual keycode. It's not a bug in the browser or OS.

yanfali:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values if you check the docs, unidentified is mostly associated with old browser versions. You are missing the point. You would get an event if it hadn't recognized, therefore it isn't Unidentified.

yanfali:
I think I understand where the disconnect is. Your assumption, I think, is that you assume we look at all input codes. We do not. We only bind to specific input codes. If we just looked at all input, then unidentified would need to be dealt with. In our case, we should never see unidentified because keypress.js had to identify it in order to fire the event. Therefore logically it should never happen in this piece of code. No point writing code for something that either we don't care about, or shouldn't happen as we have a safe default.

yanfali:
What your new code does, is reintepret the code if it has an alternative mapping. So for example. When I tested your code, it actually through I was hitting LGUI on OSX. So what happens now is LGUI fires, but a more specific test is applied and it see's it actually ContextMenu and uses the correct code. Does that make more sense now?

fauxpark:
> if you check the docs, unidentified is mostly associated with old browser versions

That is fair, I realise that this particular KeyboardEvent is probably very rare.

> Your assumption, I think, is that you assume we look at all input codes. We do not. We only bind to specific input codes. If we just looked at all input, then unidentified would need to be dealt with. In our case, we should never see unidentified because keypress.js had to identify it in order to fire the event.

This makes more sense now, I saw but did not quite parse the implication of `generateKeypressCombos()`. So the possibility of Unidentified is then further limited to only the ContextMenu key on Safari, and thus not worth caring about.

> What your new code does, is reintepret the code if it has an alternative mapping. So for example. When I tested your code, it actually through I was hitting LGUI on OSX. So what happens now is LGUI fires, but a more specific test is applied and it see's it actually ContextMenu and uses the correct code. Does that make more sense now?

Yes, I know how it works. That's why I initially placed the check in the modifier handling switch case, because Keypress.js is not *entirely* wrong when it claims that keyCode 93 is LGUI. Only mostly. But your solution is indeed much better.

yanfali:
Let me know when you happy with this and we'll get it merged. Thanks for working on this @fauxpark

fauxpark:
I'm happy if you're happy.

Issue #287, opened 2019-03-17
Suggestion: new layers should only be saved if something is added to it


1hella:
I had the problem of a bunch of empty layers being created by clicking on them, which in my case gave an error during compilation.

I think just clicking a layer shouldn't save it. I suggest that if somebody clicks on a new layer, it should only be saved after a key is changed on it. In other words, when clicking away from a layer, if every key code is KC_NO, delete it.

Alternatively, having a button to delete an accidentally created layer would work as well.

yanfali:
Yep thanks for raising issue. It's been on my list to add layer editing for a while so I will be working on this in the future.

yanfali:
@1hella give the latest configurator a try. It should be working the way you expect. Thanks for the suggestion. We also support sparse layers, but the downside is the keymap generated will be larger, so this may cause problems for some keyboards.

PR #305, opened 2019-04-07
Issue #287: handle empty layers gracefully

yanfali:
- clear empty layers if there are no active keycodes
 - export sparse keymaps with empty layers when there are higher
   numbered layers
 - send compiler dummy KC_TRNS intermediate layers when there are empty
   layers with a higher layer above it.
 - ignore empty layers when exporting keymap
 - refactor code and give variables and functions more approrpriate names

### DIFF
--- a/src/store/modules/keymap.js
+++ b/src/store/modules/keymap.js
@@ -52,30 +52,61 @@ const getters = {
     return size(state.keymap[_layer]);
   },
   isDirty: state => state.dirty,
+  /**
+   * Convert internal representation of keycodes into QMK keycodes for export or compilation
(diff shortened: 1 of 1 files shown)


noroadsleft:
> - ignore empty layers when exporting keymap

I'm not sure that part is working as intended; I still get empty layers in both the exported JSON and the `keymap.c` after compiling. The compile job completes successfully though.

yanfali:
You should get empty layers if there's a valid layer about it. It elides empty layers only if there is no valid layer above it. We have to do that to preserve the position in the array.

noroadsleft:
Ahh, okay. Got it.

:+1:

PR #307, opened 2019-04-13
Add default for Romac

stanrc85:
Add default to match default keymap in QMK for new Romac macro board.


yanfali:
1 comment, otherwise LGTM

stanrc85:
> Change the keymap name from default, backend will fail if the keymap name conflicts with an existing keymap. Anything that isn't checked into the keymaps dir will work.

Fixed!

yanfali:
Awesome. Thanks

PR #318, opened 2019-04-23
add treadstone48 default keymap

marksard:
I created and compiled from https://config.qmk.fm/, written to treadstone48, after that confirmed with behavior.

Best regards,


yanfali:
Awesome. Thanks.

Issue #331, opened 2019-05-04
white list characters in keymap name


yanfali:
To prevent compilation errors for special characters

PR #332, opened 2019-05-05
Whitelist the characters we allow in keymap names

yanfali:
- prevent people sending invalid keymap names to the API or saving to the disk

Issue #331 

All characters not in a-z0-9_ are removed from filenames saved to disk and sent to API. I do not prevent users from inputting them. I just ignore them.


noroadsleft:
I think `-` should be added to the white list, but other than that, LGTM.

PR #340, opened 2019-05-08
Colorway updates

noroadsleft:
Updates the architecture so all the colorways are in individual files, and also adds some functionality that hopefully should make adding new colorways more straightforward.


yanfali:
Does it make sense to create GMK SA and DSA directories?

yanfali:
Couple of tiny linting issues but otherwise LGTM.

noroadsleft:
> Does it make sense to create GMK SA and DSA directories?

That's a really good idea, actually.

PR #341, opened 2019-05-09
Index colorways by directory

yanfali:
- reduces import noise at top level
 - hides complexity of imports in directory

### DIFF
--- a/src/components/colorways/index.js
+++ b/src/components/colorways/index.js
@@ -1,54 +1,13 @@
 import dsa_galaxy_class from './dsa/galaxy_class';
-import sa_carbon from './sa/carbon';
-import sa_danger_zone from './sa/danger_zone';
-import sa_jukebox from './sa/jukebox';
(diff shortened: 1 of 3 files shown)


src/components/colorways/index.js
```diff
@@ -1,54 +1,13 @@
...
-    gmk_nautilus,
-    gmk_olivetti,
-    gmk_olivia,
-    gmk_serika,
-    gmk_space_cadet,
-    gmk_striker,
-    gmk_ta_royal_alpha,
-    gmk_wob
-  ],
+  list: [dsa_galaxy_class, ...gmk, ...sa],
```
noroadsleft:
```suggestion
  list: [dsa_galaxy_class, ...sa, ...gmk],
```

Makes it so the ordering stays the same in the UI.

yanfali: oh, fair enough. I was going for alpha

PR #349, opened 2019-05-16
Scrolls the selected keyboard back into view

XScorpion2:


yanfali:
It's almost there, the problem is keybindings are a bit messed up after.

XScorpion2:
ya, pushed it up just so it is tracking as an improvement. Still going to poke at it some more.

PR #351, opened 2019-05-17
Drop down remembers who's the boss!

XScorpion2:


XScorpion2:
I mean we could search the keyboards list ourselves for the element index to set the typeAheadPointer, but would still need to use var offsetTop = active.offsetTop - active.offsetHeight; for the scrollTo code.

yanfali:
Yeah I'm ok with the don't manipulation. Use a filter on the original list to find the index, not the float math thing. I know it's slower but it's clearer what the intent is.

yanfali:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

XScorpion2:
right on, will fix tonight.

yanfali:
Also, you can delete focus() if we're not using it anymore. Thanks again.

XScorpion2:
focus() is still used by the other drop downs. 
will fix the linting issues

noroadsleft:
@XScorpion2 This looks great. In testing it seems that if you select a keyboard and then open the list again, the list is scrolled one position up from the current keyboard (e.g. selecting `clueboard/66/rev3` and then opening the list shows `clueboard/66/rev2` at the top, though `rev3` is highlighted).

Minor nitpick but this is already several orders of magnitude better than current master. Might be my browser rendering/settings or something. If it stays like this I can live with it.

:heart:

XScorpion2:
@noroadsleft that behavior was intended as then you can easily see and select the previous keyboard

noroadsleft:
that behavior was intended as then you can easily see and select the previous keyboard

:+1:

PR #356, opened 2019-05-22
Make 1U keymaps great again

yanfali:
Improves readability of keycodes in 1.25U and smaller keys on ergo and ortholinear.

 - force line breaks for legends with breaks

Screen Shot 2019-05-21 at 22 50 38

 - force line breaks for raw keycodes

Screen Shot 2019-05-21 at 22 50 46

### DIFF
--- a/src/components/BaseKey.vue
+++ b/src/components/BaseKey.vue
@@ -201,6 +201,10 @@ export default {
       );
     },
     formatName(name) {
+      if (this.u < 1.75) {
+        name = name.replace(' ', '\n');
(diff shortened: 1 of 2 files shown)


noroadsleft:
`LT()` and Hold/Tap keys (`[LR](Ctl|Sft|Alt|GUI)`) look wonky with this change applied, but I love the idea.

`KC_MINS` also gets strange.

image

yanfali:
Oh good catch. I'll move it lower to the PrintKey only.

yanfali:
@noroadsleft this should work better

PR #370, opened 2019-05-24
Translate Print Keymap button to French

noroadsleft:
It didn't exist when we started translations.

Request review from @gab-a because I don't actually speak French.

### DIFF
--- a/src/i18n/fr/index.js
+++ b/src/i18n/fr/index.js
@@ -39,8 +39,8 @@ export default {
         title: 'Importer un fichier .JSON de disposition des touches'
       },
       printKeymap: {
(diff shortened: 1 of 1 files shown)


gab-a:
for title I would conjugate the verb 'imprimer' because I did this for every other title
and for the label I would add the pronoun 'la' otherwise it sound weird in french, we use pronouns a lot :D but it makes it even longueur. So if needed 'imprimer' could be shorten to 'imp.' and most french would understand it is not uncommon.
I would go with this 
`title: 'Imprime les strates de la disposition',
label: 'Imprimer la disposition'`

PR #371, opened 2019-05-25
support number pad and function keys #298

### LINKED ISSUE #298: Support number pad and function keys
A few suggestions:
- Support number pad keys using `event.location` as you already did with left/right modifiers https://github.com/qmk/qmk_configurator/commit/55f6df70bea9ef7a7a5dc1165d73a7966cd4adca.
- Support function keys:
  - Add the `event.key` value to `/modules/keycodes/ansi.js` (e.g.  `keys: "F1"`);
  - Intercept the browser function keys (i.e. https://keycode.info).

Thank you.

mukealicious:
- detects number pad keys following the pattern used for mods in https://github.com/qmk/qmk_configurator/commit/55f6df70bea9ef7a7a5dc1165d73a7966cd4adca
- adds function row support, and prevents default behavior using `preventDefault` (only on function keys)

closes #298


src/store/modules/keycodes/ansi.js
```diff
@@ -3,20 +3,20 @@ export default [
 
   { name: 'Esc', code: 'KC_ESC', keys: 'esc' },
   { width: 1000 },
-  { name: 'F1', code: 'KC_F1' },
-  { name: 'F2', code: 'KC_F2' },
-  { name: 'F3', code: 'KC_F3' },
-  { name: 'F4', code: 'KC_F4' },
+  { name: 'F1', code: 'KC_F1', keys: 'f1' },
```
fauxpark: Could you also do the same for iso-jis.js?

fauxpark:
Seems like the Configurator can already recognise the numpad numbers on Firefox on Win10 & macOS, and Chrome on Windows.

Win10 Firefox doesn't recognise `KC_PDOT`. None of them recognise `KC_PENT` or the function row.

Apple keyboards also have `KC_PEQL` and function keys going up to F19 - I would add the full range up to F24.

mukealicious:
It appears the third party library we're using here doesn't support anything above f12.

https://github.com/dmauro/Keypress/blob/master/keypress.coffee#L884-L895

We're currently limited to their keycode dictionary, and overriding in certain cases based on `event.DOM_KEY_LOCATION`.

Digging a little bit more through MDN, it appears `KeyboardEvent.keyCode` is deprecated, and they are now recommending `KeyboardEvent.key` as the standard. ([source](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Properties))

Here's a [demo](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Result) using `KeyboardEvent.key`

I think working directly with the browser API would be a better long term solution than keypress.js. It would still require `DOM_KEY_LOCATION` to infer things like right modifiers and numpad, as we're doing now.

fauxpark:
I think that the keyLUT with the ContextMenu key bypasses Keypress.js (by indeed using ev.key). Maybe try adding to that?

mukealicious:
I tried, but the problem is that the listener comes from keypress.js, which is why we are limited to their dictionary.

https://github.com/qmk/qmk_configurator/blob/master/src/jquery.js#L19

No events are generated for keys like F13, for example, which I have been able to successfully log in the MDN `KeyboardEvent.key` demo.

fauxpark:
That's a shame. Oh well, I doubt many people will want to assign F keys above 12 anyway.

yanfali:
I'm getting really odd behavior in FF, none of the numpad numeric keys are registering. It's very inconsistent. Debugging

yanfali:
https://github.com/dmauro/Keypress/blob/master/keypress.coffee#L728-864 we should probably use this list in the keys bindings.

yanfali:
We could try sending a PR to extend the list to keypress.js. I'm using it because it normalizes browsers for me, which is pretty valuable in of itself. Plus it allows cool combo keys.

yanfali:
Actually thinking about this more. I think we ignore their bindings for numpad, and simply do the key detection work around to get firefox support. It seems like the library is a bit wiggy about numpad on FF

yanfali:
OK, false alarm. PEBCAK. Vimium was eating my keypresses. This looks fine on FF. Sigh

PR #374, opened 2019-05-25
Support shifted keycodes as input

yanfali:
- add a separate handler for mods
   - detect keyup event rather than keydown so now we can support
     shifted keypresses directly. Idea from community.

### DIFF
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -106,6 +106,8 @@ const functionKeys = [
 // Used exclusively to detect mods on Keyup so we can support modded input
 function keyupHandler(meta, ev) {
   let _meta = meta;
+
+  // detect left and right mods
   if (ev.location === ev.DOM_KEY_LOCATION_RIGHT) {
     _meta = store.getters['keycodes/lookupKeycode'](mods[meta.code]);
   }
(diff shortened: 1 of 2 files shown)


noroadsleft:
Two things:

## 1.

With this patch applied, tapping the Context Menu key `KC_APP` assigns `KC_LGUI` to the selected key. The current live version of Configurator correctly assigns `KC_APP`.

## 2.

This may be (probably is) a browser issue rather than code, but with this patch Firefox requires two taps of the Alt key (either one) to assign their keycodes by typing. The first activates the browser's menu (File, Edit, View, etc.) on keyup, and then the second disables the menu on keydown, and then the keyup assigns the keycode, like so:

Keypress | Event | Result
:-: | :-: | :--
1 | Down | nothing
1 | Up | Browser menubar activates
2 | Down | Browser menubar deactivates
2 | Up | Keycode is assigned

Current Configurator does the following with the Alt key (Firefox 66.0.5, 32-bit on Windows 8.1):

Keypress | Event | Result
:-: | :-: | :--
1 | Down | assigns the keycode
1 | Up | Browser menubar activates
2 | Down | Browser menubar deactivates
2 | Up | nothing


Brave Browser (v0.64.77 Chromium: 74.0.3729.169 64-bit) doesn't activate the menu based on the Alt key. (In fact, it doesn't have a traditional menubar like Firefox does, just a hamburger menu. I imagine Chrome is the same.) It just assigns the code based on keydown (Live) or keyup (this patch).

noroadsleft:
Ok, `KC_APP` and Alt key issues are both resolved, as best as I figure they can be. Firefox and/or Windows always seems to respond to certain keys no matter what.

PR #382, opened 2019-06-02
Make content keys selectable

yanfali:
- this enables keyboard input, making it consistent with other keys we
   support. This should enable LT and all MT variants we support

### DIFF
--- a/src/store/modules/keymap.js
+++ b/src/store/modules/keymap.js
@@ -177,8 +177,10 @@ const mutations = {
   },
   setSelectedContent(state, index) {
     if (state.selectedIndex === index) {
+      // key already selected - toggle inner content selection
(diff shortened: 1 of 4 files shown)


src/store/modules/keymap.js
```diff
@@ -172,32 +173,55 @@ const actions = {
 const mutations = {
   setSelected(state, index) {
     state.selectedIndex = index;
+    state.selectedContent = false;
+  },
+  setSelectedContent(state, index) {
+    if (state.selectedIndex === index) {
+      // key already selected - toggle inner content selection
+      state.selectedContent = state.selectedContent ? false : true;
```
yanfali: Should just be !

yanfali:
@mechmerlin https://test.config.qmk.fm for a live version

Issue #368, opened 2019-05-24
add i18n to Print view


yanfali:
i18n was not done for the Print.vue route

PR #383, opened 2019-06-02
I18n print

zekth:
fixes: https://github.com/qmk/qmk_configurator/issues/368

also i may have some suggestions about translations to make it easier to maintain and contribute. Here is an example: https://github.com/reef-pi/reef-pi/blob/master/docs/internationalization_contributing.md

### DIFF
--- a/src/views/Print.vue
+++ b/src/views/Print.vue
@@ -3,52 +3,52 @@
     <div class="print-controls">
       <button id="leavePrint" @click="gohome">
         <font-awesome-icon icon="chevron-left" size="lg" fixed-width />
-        Back
+        {{$t('message.back.title')}}
       </button>
       <button id="leavePrint" @click="print()">
         <font-awesome-icon icon="print" size="lg" fixed-width />
-        Print
+        {{$t('message.print.title')}}
       </button>
     </div>
     <div class="meta-info">
       <table>
         <tr>
-          <th>Keyboard</th>
+          <th>{{$t('message.keyboard.label') }}</th>
           <td>{{ keyboard }}</td>
         </tr>
         <tr>
-          <th>Layout</th>
+          <th>{{$t('message.layout.label') }}</th>
           <td>{{ layout }}</td>
         </tr>
         <tr>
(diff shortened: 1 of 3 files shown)


src/views/Print.vue
```diff
@@ -117,10 +119,10 @@ export default {
     },
     print() {
       if (this._notes === '') {
-        this._notes = 'My awesome keymap';
+        this._notes = this.$t('message.layer.empty');
```
noroadsleft: Should be `message.notes.empty` here.

```suggestion
        this._notes = this.$t('message.notes.empty');
```

noroadsleft:
Firefox seems to use the `placeholder` attribute values for printing if the user doesn't set their own. I'm okay with this though.

image

zekth:
> Firefox seems to use the `placeholder` attribute values for printing if the user doesn't set their own. I'm okay with this though.
> 
> image

This seems unrelated with my changes but didn't know FF uses placeholder value if empty

yanfali:
@zekth if you print, those go away and are replaced with stock text.

yanfali:
I think this is great. My plan was however to leverage ES6 imports and simply put this into it's own file.  And include it into index.js much like potato.js. This is really it's own separate application that leverages the same data store. Can we create a new file with these keys called `print.js`, import that and then include it using `...print`. Let's also namespace this into it's only subtree like potato:

```JSON
export default {
  print: {
     ...rest of keys here
  }
}
```

I'm about to add new part of the application to test keypresses, and I was planning on localizing it using this scheme. Let me know what you think. The CSV idea is interesting, I will think about it.

zekth:
Internationalization is really not a fun part for any developper. Using CSV is pretty straight forward and easy to read / edit. Using Js fles could be long. If you want i can address that.

Also having multiple files for namespaces has no real benefits as there is not so much entries. Or if you prefer `index.js` would just refers all the namespaces, so we would need to add something like `common.js`

yanfali:
This looks good. Thank you. I will test it today.

PR #387, opened 2019-06-03
fixing translation typo

zekth:


noroadsleft:
I actually spotted this semi-recently in a wade through Google Translate, but I don't actually speak French so couldn't determine whether or not this was better/correct.

zekth:
> 
> 
> I actually spotted this semi-recently in a wade through Google Translate, but I don't actually speak French so couldn't determine whether or not this was better/correct.

Added a comment on the french translation commit tho.

noroadsleft:
Must have missed it, then. I haven't had write access here for long, and I'm still working out my GitHub Notification settings.

Issue #342, opened 2019-05-10
built in keycode tester


yanfali:
use something like the at101 blackheart layout and a fixed keymap. As keypresses come in, animate the key that was activated. And change it's color ala EKSH. Allow reset. display whatever raw info we have below from browser.

PR #390, opened 2019-06-04
Integrated Keyboard Testing

yanfali:
Because Elite Keys website has gone dark, I felt like there was a need for a keyboard tester in configurator. This makes it relatively OS independent and while it doesn't do chatter testing, does provide the basics, so a user who has flashed their keymap can do a basic sanity test of the codes being sent to the browser.

Issue #342


public/style.css
```diff
@@ -586,7 +586,10 @@ a img {
 #import,
 #settings,
 #printkeymaps,
+#testkeys,
 #leavePrint,
+#leaveTest,
+#resetTest,
```
zekth: why using ID target and not just a class for this?

yanfali: This is mostly a legacy issue, the style sheet was done with IDs instead of classes, so it's inertia. We're going to redesign the UI, so I'm not too concerned, but it's definitely an anti-pattern these days.

yanfali:
Deployed for testing at https://test.config.qmk.fm

zekth:
Testing is ok an pretty easy to use.
The only concern i have ATM is the other layouts like ISO (oh no!).

So here i'm on an AZERTY keyboard, and when i press `A` i got this:
`KEY-DOWN -- QMK: 'KC_Q' Event key: 'a', Code: KeyQ, keyCode: 65`
So event is handled correctly, but the key showing on the layout is `Q`, which is wrong for me as user side, but on the matrix side it's correct.
This could confuse beginners and people who wants to test their keyboard when they use other layouts than standard `ANSI`. A nice feature would be to show the interpreted key by the OS on the UI; then the user will know which key on the matrix is pressed and what is interpreted by the OS.

yanfali:
This is a good Idea. I have a couple of ideas on how to make that more obvious.

yanfali:
Screen Shot 2019-06-04 at 20 14 51

OK, this is what I came up with

drashna:
@yanfali could we get hex numbers on there too? 

Also, colemak/dvorak/etc support?

yanfali:
Screen Shot 2019-06-04 at 21 09 18

yanfali:
@drashna probably not in version 1. We're discussing how to do ISO too

yanfali:
@drashna I added a hex toggle for you. it's an easter egg. find it.

src/components/VisualTesterKeymap.vue
```diff
@@ -0,0 +1,273 @@
...
+    keydown(ev) {
+      if (ev.repeat) {
+        return;
+      }
+      this.timing[ev.code] = performance.now();
+      ev.preventDefault();
+      ev.stopPropagation();
+      const pos = this.codeToPosition[this.firefoxKeys(ev.code)];
+      this.writeToStatus(
+        `KEY-DOWN -- QMK: '${this.getQMKCode(pos)}' ${this.formatKeyEvent(ev)}`
```
zekth: This can be exported in a `formatPrompt` function don't you think?

zekth:
2 remarks:
1. on narrow screen it needs a bit of margin for the buttons:
2019-06-05 11_14_21-QMK Configurator

2. Addition of detected keycodes is nice. Maybe adding a description in the bottom of the page explaining what are they and why in case of AZERTY layout for example keycode is not the same as key.

Otherwise it's shippable to me.

src/components/VisualTesterKeymap.vue
yanfali: yeah, good idea. Sometimes I don't do this until I hit the 3rd use.

yanfali:
Yah I knew about the button overflow problem. I'll have to think about it. I may remove the text when you reach a certain size.

yanfali:
Screen Shot 2019-06-05 at 15 06 40

CSS removes text when the browser gets smaller. They are not critical functions and the icons are pretty good.

src/components/VisualTesterKeymap.vue
yanfali: I looked into it and they're different enough it doesn't really help, it just makes it harder to figure out what's going on, so I'm going to leave it for now.

PR #395, opened 2019-06-05
[Colorway] GMK Plum

zekth:
See:
2019-06-05 14_47_27-QMK Configurator


yanfali:
@mechmerlin I think you got permission from Jessica right?

zekth:
Daily driver man <3 a pleasure

PR #400, opened 2019-06-06
fix browser detection

zekth:
A bit odd detection but works. Tried on severals navigators.


zekth:
Note: was testing using different browser and using this extension: https://addons.mozilla.org/fr/firefox/addon/user-agent-switcher-revived/

zekth:
Fixed sorry.

yanfali:
Thanks @noroadsleft and nice work @zekth

Issue #361, opened 2019-05-23
Load keymap.JSON from URL


yanfali:
point it at a raw git URL and have it make a JSON request for the data, not sure it'll run afoul of CORS or not.

PR #403, opened 2019-06-07
add support to load keymaps from url

zekth:
This address: https://github.com/qmk/qmk_configurator/issues/361

We may need to add a modal to the project because we still use old school alerts / prompt in the project.

### DIFF
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -1,5 +1,24 @@
 <template>
   <div id="controller-bottom" class="botctrl">
+    <Veil :isVisible="isVeilOpened">
+      <template #contents>
+        <div class="input-url-modal">
+          <div>
+            <label for="url-import-field">Url:</label>
+            <input
+              ref="urlimport"
+              id="url-import-field"
+              type="text"
+              v-model="urlImport"
+            />
+          </div>
+          <div>
+            <button @click="closeVeil">cancel</button>
+            <button @click="importUrlkeymap">Load</button>
+          </div>
+        </div>
+      </template>
+    </Veil>
     <div class="botctrl-1-1">
       <button
         class="fixed-size"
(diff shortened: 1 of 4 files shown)


yanfali:
Yeah, we have a generic Veil component, so adding a Modal should be pretty easy.

yanfali:
@zekth you want to take a stab at it?

zekth:
> 
> 
> @zekth you want to take a stab at it?

yup

src/components/ControllerBottom.vue
```diff
@@ -357,23 +411,33 @@ export default {
...
   }
 };
 </script>
 <style scoped>
+.input-url-modal {
+  background-color: #eee;
+  border-color: #ccc;
+  border-radius: 5px;
+  width: 400px;
+  min-height: 400px;
```
noroadsleft: Is this supposed to be `min-width`?

src/i18n/en/index.js
```diff
@@ -40,6 +40,9 @@ export default {
       importJSON: {
         title: 'Import QMK Keymap JSON file'
       },
+      importUrlJSON: {
+        title: 'Import QMK Keymap JSON from Url'
```
noroadsleft: `URL` should always be capitalized.

```suggestion
        title: 'Import QMK Keymap JSON from URL'
```

yanfali: Yeah, we left it unstyled because zekth isn't big on CSS. You want to take a stab at it @noroadsleft ? I was going to look at this weekend, but this is something right up your alley. I may write a better modal at a future date that can be re-used, but this is good enough for now mostly likely.

zekth:
Just styling remaining yap.

Also a nice feature would be to use a query string also to load it like:
`https://config.qmk.fm/?load=http://foo.bar`
And by this way we can make something like `Share keymap` using tinyUrl. Your thoughts?

src/i18n/en/index.js
noroadsleft: Yeah, I've already got some ideas as far as CSS. It's not a lot of rules but it makes a big improvement visually IMO.

zekth:
Merged and fixed a little css stuff. Also we might need to fix the controller bottom grid in a further PR. Example:
2019-06-09 16_26_35-Window

noroadsleft:
@zekth Agreed, but this is good for now IMO.

PR #406, opened 2019-06-10
[tester] adjust loggin

zekth:
I've adjusted the loggin in the console to make it more readable:
From this:
2019-06-10 14_29_08-QMK Configurator
to this:
2019-06-10 14_51_44-QMK Configurator

Note: added `spellsheck=false` to avoid the browser to try to parse content and having red highlighting depends on the browser lang.

### DIFF
--- a/src/components/VisualTesterKeymap.vue
+++ b/src/components/VisualTesterKeymap.vue
@@ -38,6 +38,7 @@
         ref="status"
         cols="120"
         rows="5"
+        spellcheck="false"
         v-model="status"
       />
     </div>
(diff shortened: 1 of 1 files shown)


src/components/VisualTesterKeymap.vue
```diff
@@ -32,14 +32,14 @@
...
+      <div
         class="status-log"
         id="terminal"
         ref="status"
-        cols="120"
-        rows="5"
-        v-model="status"
-      />
+        spellcheck="false"
+        style="width:869px;text-align:left;"
```
yanfali: let's extract this to <style>

zekth: this needs some refactor of css stylesheet because it uses also the `#terminal` like homepage.

src/components/VisualTesterKeymap.vue
```diff
@@ -181,7 +190,13 @@ export default {
         msg.push(`in ${(endTS - this.timing[ev.code]).toFixed(3)}ms`);
       }
       msg.unshift(
-        `Event key: '${ev.key}', Code: ${ev.code}, keyCode: ${ev.keyCode}`
+        `Event key: <span class="log-green">${ev.key.padEnd(
```
yanfali: let's make this a function since we're doing it so much:
```javascript
function greenMarkup(text, padlen) {
  return `<span class="log-green">${text.padEnd(padlen, ' ')}</span>`;
}
```

yanfali: function name is only a suggestion, pick something that makes more sense if you don't like it.

yanfali:
Ah, true. I forgot it was an ID.

src/components/VisualTesterKeymap.vue
```diff
@@ -181,7 +191,14 @@ export default {
         msg.push(`in ${(endTS - this.timing[ev.code]).toFixed(3)}ms`);
       }
       msg.unshift(
-        `Event key: '${ev.key}', Code: ${ev.code}, keyCode: ${ev.keyCode}`
+        [
```
yanfali: Subrashi!

PR #409, opened 2019-06-11
[tester] Add ISO layout

zekth:
Added ISO layout:
2019-06-12 15_13_36-QMK Configurator

Some labels on keys are wrong because of a problem in keycodes handling:
https://github.com/qmk/qmk_configurator/blob/master/src/store/modules/keycodes/index.js#L8

This is wrong, because there is several times the same keycode due to the different layout and spread operators. Therefore the results of this method are wrong: https://github.com/qmk/qmk_configurator/blob/master/src/store/modules/keycodes/index.js#L15

It works for ANSI, because it's the first keycode array to be spread. This need to be refactored in another PR.


yanfali:
Hmm, maybe we should just introduce a layout for ISO users? I mean the code does register when you hit it in the debug right? I think it's just odd for it to hang out like that for ANSI users, since it's really KC_NUBS not KC_BSLS

zekth:
> ust introduce a layout for ISO users? I mean the code does register when you hit it in the debug right? I think it's just odd for it to hang out like that for ANSI users, since it's really KC_NUBS not KC_BSLS

it was just to not have to setup a ISO layout for the tester. But you're right.

zekth:
Edited PR. Spotted some problems. Need your opinion.

yanfali:
OK, this is looking promising. I think let's get rid of the dropdown, I think we'll only have three choices, So maybe radio buttons? Put in JIS but disable it.

Let's not default to ISO obviously. Let's steal the language detection code into a file and re-use it.

I think it looks really promising so far. After you get this working we can add in JIS and detect based on country.

src/components/VisualTesterKeymap.vue
```diff
@@ -62,18 +73,19 @@ export default {
   name: 'visual-tester-keymap',
   extends: BaseKeymap,
   async mounted() {
+    this.availableLayouts = Object.keys(this.layouts).sort();
     document.addEventListener('keydown', this.keydown);
     document.addEventListener('keyup', this.keyup);
     await this.init();
-    this.setSize(this.calculateMax(this.layout));
+    this.setSize(this.calculateMaxFromLayout(this.layouts[this.layout].layout));
```
yanfali: ok, that's kind of a mess.

src/components/BaseKeymap.vue
```diff
@@ -45,6 +44,13 @@ export default {
         max.y *= this.config.SCALE;
       }
       return max;
+    },
+    calculateMaxFromLayout(layout) {
+      return this._calculateMax(layout);
+    },
+    calculateMax(newLayout) {
```
yanfali: not sure about this. just leave it in the tester keymap version. It has no value in the base keymap

src/store/modules/tester/codeToPos.js
```diff
@@ -0,0 +1,213 @@
+export const ANSI = [
```
yanfali: I think they differ by 2 positions. Let's just take ANSI and slice it to make ISO?

yanfali: I think they differ by 2 positions. Let's just use ANSI as the base and slice ISO

src/store/modules/tester/keymap.js
```diff
@@ -0,0 +1,214 @@
+export const ISO = [
```
yanfali: ditto comment on slicing

src/store/modules/tester.js
```diff
@@ -1,369 +1,288 @@
...
-      { label: 'Win', x: 11.25, y: 5.5, w: 1.25 },
-      { label: 'Menu', x: 12.5, y: 5.5, w: 1.25 },
-      { label: 'Ctrl', x: 13.75, y: 5.5, w: 1.25 },
-      { label: '\u2190', x: 15.25, y: 5.5 },
-      { label: '\u2193', x: 16.25, y: 5.5 },
-      { label: '\u2192', x: 17.25, y: 5.5 },
-      { label: '0', x: 18.5, y: 5.5, w: 2 },
-      { label: '.', x: 20.5, y: 5.5 }
-    ]
+    ISO: {
```
yanfali: yeah, let's base one off the other and slice.

src/store/modules/tester.js
```diff
@@ -1,369 +1,288 @@
...
+  setLayout(state, layout) {
+    state.layout = layout;
+  },
   setKeymap(state, keymap) {
     state.keymap = keymap;
   },
-  setActive(state, pos) {
-    Vue.set(state.keymap[0][pos], 'active', true);
-    Vue.set(state.keymap[0][pos], 'detected', false);
+  setActive(state, { keymapIdx, pos }) {
```
yanfali: I don't like this. It's too messy. We should hide the layout abstraction inside the store, not the component. The store should know what layout is currently selected and provide getters. This just exposes too much inner implementation.

yanfali:
OK, I've done some code review. I think it's great the progress you've made, but I think we're leaking too much implementation into the component. I would prefer to encapsulate more  of the knowledge of what layout we are actually rendering into the store where it belongs. Having it in both in the component and the store is giving me a headache.

src/store/modules/tester/codeToPos.js
zekth: I don't agree with this. Slicing would be a real mess to read. Like having the base (ANSI) layout and manipulate it to set it to ISO or JIS? Not maintainable IMO.

zekth:
ATM the tester is not really well prepared for handling several layouts. Struggling with keymaps / codeposition gave me a headhache too. IMO it's not maintainable.

I'd personnally make a simple layout object (ANSI/ISO/JIS) with everything (keycode, label, position and so on) in it and not split it throught several objects.

Having implementation into the component is not a problem as it's not meant to be shared because it concerns only this component and will not be shared to others.

NOTE: ISO default layout was a mistake i've forgotten.

yanfali:
@zekth we should talk on discord, but I'll make a PR against your PR once I've had a chance to play around. Trust me, the problem isn't any different from multiple layouts for a keyboard.

zekth:
> @zekth we should talk on discord, but I'll make a PR against your PR once I've had a chance to play around. Trust me, the problem isn't any different from multiple layouts for a keyboard.

Make the modifications you want as i'm a bit lost in all those implementations...

zekth:
Added radio buttons, fixed default layout detection, linting. Shippable for me.
May you add some mode comments in the keymap part @yanfali  ?

yanfali:
@zekth let's get rid of the label LAYOUT, and move the ANSI and ISO radios to the center of the page. @noroadsleft any other suggestions?

yanfali:
I'll send a PR to your branch about the split stuff

src/store/modules/tester/codeToPos.js
```diff
@@ -0,0 +1,142 @@
...
+  'F4',
+  'F5',
+  'F6',
+  'F7',
+  'F8',
+  'F9',
+  'F10',
+  'F11',
+  'F12',
+  'F13',
```
yulei, line 15: Should it is "PrintScreen" ?
(C0K5PP3VFWM1`0J1_ZS%LT

yanfali: I'm sorry, but Event key and Code are generated by the Browser, not the web app. Those are the reported values from here https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
That's not generated by QMK at all. If you look at QMK it says KC_PSCR, which is print screen, but print screen is windows specific I believe and is basically F13 which is the original code that operating systems remapped to be print screen. So it is actually F13, not print screen. That's just what your operating system interprets it as. It's in the HID spec.

yanfali:
Here's the code that generates F13 output:

```javascript
 formatKeyEvent(ev, time) {
      const msg = [];
      if (time) {
        msg.push(`in ${time}ms`);
      }
      msg.unshift(
        [
          'Event key:',
          this.greenMarkup(ev.key, 11),
          'Code:',
          this.greenMarkup(ev.code, 13),
          'KeyCode:',
          ev.keyCode
        ].join(' ')
      );
      return msg.join(' ');
    },
```

As you can see there's no indexing or look ups. It's printing the event key, code and keycode values directly to a string.

yulei:
It should display as QMK:KC_F13, not QMK:KC_PSCR, this was a misleading. Yes, the PSCR down event will be handled by the browser.
I59@@6{IQO3{3IR~@02D`2I

yulei:
And should not highlight the PSCR button while F13 was pressed.

yanfali:
Oh I see. Well thanks for clearing that up for me. It would be helpful if
you were more explicit about bug reports as otherwise the reader is left to
guess what your issue is since those keys are so rarely used that it
doesn't get tested very well. Submit a PR if you know how to fix it. Thanks
👍

yulei:
ok, i'll try to fix it.  Seems https://www.zfrontier.com/lab/keyboardTester  has the same issue as this, thus they should copy those codes.

yanfali:
I don't have a keyboard with that many keys right now. I'll probably have
to program a layer to test this too. Thanks for your help.

yanfali:
@yulei are you on macOS? I seem to be getting this behavior on Mac. And I think it's correct since this is what macOS is reporting. Can you try it on windows?

Screen Shot 2021-09-28 at 20 15 12

My keymap is set up like this. So I'm not sure it's an actual bug more than a quirk of how macOS treats print screen.

Screen Shot 2021-09-28 at 20 16 33

yulei:
I'm currently using Win10. Yes, we know that F13 under MAC was same as PrintScreen, but I don't think the Tester should convert F13 to PSCR because this was not true under Windows. Please check #998

Issue #222, opened 2018-12-03
experiment with hiding the console my default and expanding it when the user hits compile.


yanfali:
The console may be a little confusing for new users. Only show it when they hit compile, otherwise have it collapsed by default.

yanfali:
Resolved by #583

PR #410, opened 2019-06-15
terminal hiding

zekth:
adress: https://github.com/qmk/qmk_configurator/issues/222

### DIFF
--- a/src/components/StatusPanel.vue
+++ b/src/components/StatusPanel.vue
@@ -1,15 +1,19 @@
 <template>
   <div id="status">
-    <div id="terminal" v-on:click="showTerminal" v-show="!terminalVisible">
-      ^ Expand terminal ^
-    </div>
-    <textarea
-      id="terminal"
-      v-model="message"
-      ref="terminal"
-      v-show="terminalVisible"
-      readonly
-    />
+    <transition name="terminal-fade">
+      <div id="terminal" v-on:click="showTerminal" v-show="!terminalVisible">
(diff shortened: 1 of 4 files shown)


src/components/StatusPanel.vue
```diff
@@ -25,7 +31,12 @@ export default {
     }
   },
   computed: {
-    ...mapGetters('status', ['message', 'scrollToLatest'])
+    ...mapGetters('status', ['message', 'scrollToLatest']),
+    isTerminalVisible: {
+      get() {
+        return this.$store.state.app.terminalVisible;
```
yanfali: Use mapState

src/jquery.js
```diff
@@ -436,6 +436,14 @@ function enableCompileButton() {
   }, 2000);
 }
 
+function hideTerminal() {
+  store.commit('app/setTerminalVisibility', false);
```
yanfali: mapMutations

yanfali: These don't belong in here

yanfali:
OK, I just tried it out. It looks funky. I think we need a 1 line height bar in black with some text that says something like "Expand Status". We can make it clickable and it will toggle the shrinking expanding, kind of like a concertina or unfolding of an editor. What do you think?

zekth:
I can do this yes.

zekth:
@yanfali do we still want this one? If so i'll rebase it.

yanfali:
Closed in favor of #583

PR #417, opened 2019-06-23
Button to clear a layer

yanfali:
Add a control to clear a layer at a time
Screen Shot 2019-06-22 at 17 30 33

Pushing this will require confirmation and then it removes all keys from just this layer.

### DIFF
--- a/src/components/LayerControl.vue
+++ b/src/components/LayerControl.vue
@@ -13,7 +13,11 @@
         @click="clicked(layer.id)"
       >{{ layer.name }}</div>
     </div>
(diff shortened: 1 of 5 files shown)


src/components/LayerControl.vue
```diff
@@ -13,6 +13,9 @@
         @click="clicked(layer.id)"
       >{{ layer.name }}</div>
     </div>
+    <button class="clear-button" title="Clear layer" @click="clearLayer">
```
noroadsleft: This `title` attribute will need i18n support, but everything else looks good.

yanfali: good catch

Issue #184, opened 2018-08-07
Help Overlay


mechmerlin:
**Some common problems observed on the configurator** 

1. People forget to choose an appropriate layout, resulting in confusion when filling up extra "spaces" in the keycap. 

2. After keymap creation, people are unsure of where to go next. Some think that the resulting .json from `Export Keymap` is the final result. 

3. Extension of point #2, users don't understand what the word compile means and proceed to press Download Firmware which can't be clicked. 

**Request**

It would be helpful to have some kind of Help Overlay which guides users to the different parts of the configurator that needs user input. 

Step 1: Choose Keyboard
Step 2: Choose Layout
Step 3: Make Keymap
Step 4: (Optional) Export Keymap
Step 5: Compile
Step 6: Download Firmware

yanfali:
https://github.com/pulsardev/vue-tour looks promising

zekth:
Having a Tour mode seems cool tho. Vue-tour has no real updates since last year, but seeing the issues it seems ok. I can try to tackle it.

PR #422, opened 2019-07-02
[WIP] add tour

zekth:
adresses : https://github.com/qmk/qmk_configurator/issues/184

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "vue-js-toggle-button": "^1.3.2",
     "vue-router": "^3.0.1",
     "vue-select": "3",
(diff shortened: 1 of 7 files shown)


yanfali:
@noroadsleft all good points, and I've discussed it offline with @zekth I kind of like this one better, but it would need to be ported to vue as it depends on jquery https://github.com/heelhook/chardin.js

yanfali:
I think I may have found a slightly different solution

https://test.config.qmk.fm/#

Go into control panel, and toggle tutorial

yanfali:
Alternative approach #430

PR #430, opened 2019-07-07
add video tutorial to configurator

yanfali:
To help users get started, embed a video of @mechmerlin tutorial from YouTube.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -191,6 +191,10 @@ div.openSettings > button {
   bottom: 20px;
   right: 20px;
   z-index: 4000;
+  background-color: #ddd;
+  padding: 6px;
+  border-radius: 3px;
+  box-shadow: 0 0 3px #0009;
 }
 
 .help {
--- a/src/App.vue
+++ b/src/App.vue
@@ -31,10 +31,17 @@
     </footer>
     <div
       class="help"
+      :class="helpClasses"
       @click="toggleTutorial"
       :title="$t('message.help.label')"
-      @mouseenter="setMessage($t('message.help.label'))"
-      @mouseleave="setMessage('')"
+      @mouseenter="
+        setMessage($t('message.help.label'));
+        hover = true;
+      "
+      @mouseleave="
+        setMessage('');
+        hover = false;
+      "
     >
       <font-awesome-icon icon="hat-wizard" transform="rotate-22" size="3x" />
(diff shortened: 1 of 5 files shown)


yanfali:
adds an icon to the top right of UI in the shape our a wizard hat because merlin:

Screen Shot 2019-07-06 at 22 52 33

Clicking hat, brings up his youtube video tutorial

Screen Shot 2019-07-06 at 22 53 05

zekth:
Loving the summon wizard!

noroadsleft:
Some visual polish for the video player might be nice, though that doesn't necessarily need to be done here.

Also, do we want this, #422, or both?

yanfali:
> Some visual polish for the video player might be nice, though that doesn't necessarily need to be done here.
> 
> Also, do we want this, #422, or both?

I think only this one. 422 just isn't doing it for me. It doesn't feel right. Send me a PR for visual updates. I had no idea you could do that. I'm just straight embedding the video in an iframe.

noroadsleft:
> Send me a PR for visual updates. I had no idea you could do that. I'm just straight embedding the video in an iframe.

My point was you can style the `iframe` element, so it doesn't look like just a random video dropped on the page. It's minor to the eye, but makes the addition feel more polished IMO.

yanfali:
> > Send me a PR for visual updates. I had no idea you could do that. I'm just straight embedding the video in an iframe.
> 
> My point was you can style the `iframe` element, so it doesn't look like just a random video dropped on the page. It's minor to the eye, but makes the addition feel more polished IMO.

OK, send me a PR.

mateossh:
Why this specifically use `v-if` not `v-show` directive? Is this used for stopping video when iframe is hidden? Downside of unmounting component is sending the same requests every time someone toggles video.

zekth:
As there is such a few people toggling it i assume using `v-if` is more convinient, otherwise it will load the iframe in the background while you'll not even watch it.

mateossh:
Oh, I understand that approach now. I was thinking from different persperctive - if completely new user wanted to hide video and show it again, he would lose progress and new requests to YouTube would be sent.

yanfali:
It's basically don't pay the price for loading i-frame if you aren't using it. People who are using the tutorial will generally watch the entire thing, or leave it up while they play around.

PR #444, opened 2019-07-26
add angel64 default keymap 

kakunpc:
add angel64_default.json


noroadsleft:
The keymap you've submitted here doesn't match [the default keymap on qmk_firmware](https://github.com/qmk/qmk_firmware/blob/ec3954577cb080b3ea44f912a91b3c6e90e00c23/keyboards/angel64/keymaps/default/keymap.c). Were you planning on updating the keymap on qmk_firmware with these changes? We like the defaults here to match the qmk_firmware defaults as much as possible.

kakunpc:
Changed to the same default json as the qmk_firmware.

PR #455, opened 2019-08-05
Update dependencies

yanfali:
- update yarn.lock to remove bad lodash refs
 - update prettier
 - add missing eslint-plugin-prettier dep

@zekth


zekth:
LGTM

PR #457, opened 2019-08-06
Color keycode deletion links to key legend color

noroadsleft:
Makes it so the X to delete the assigned keycode is actually visible.

(This may be the most minor PR this repository has ever seen. :laughing:)

### DIFF
--- a/src/scss/themes.scss
+++ b/src/scss/themes.scss
@@ -50,7 +50,7 @@ html {
     }
 
     .key .remove {
-      color: #ccc;
+      color: inherit;
     }
     .keycode {
       background: #fff;


zekth:
It does depend on the Keyset color right?

noroadsleft:
@zekth Yes. Setting `inherit` makes the text color that's assigned to the key – which is set by keyset and key type (alpha, mod, accent) – apply to the link to delete the assigned keycode. Applying `#ccc` doesn't work all the time because some of the keysets use light colored keys. GMK Plum alpha keys for instance, end up with [this contrast](https://contrast-ratio.com/#%23cccccc-on-%23ded0af).

zekth:
LGTM so

PR #460, opened 2019-08-06
refactor of CI

zekth:
As discussed with @yanfali here is a refactor of the CI, to have more test for each PR and also get rid of legacy scripts to now use built in travis features

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,15 +4,24 @@ services:
 language: node_js
 node_js:
   - 11.12 # need this version for Jest 24, prior versions have SYMBOL bugs
-before_script:
-  - git clone --single-branch --branch config https://github.com/qmk/qmk_configurator config
-  - cp config/ga-production.js src/
+
+install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,22 +1,31 @@
...
+deploy:
+  provider: pages
+  skip_cleanup: true
+  local_dir: dist
+  github_token: $GH_TOKEN
+  keep_history: true
+  on:
+    branch: master
+
+notifications:
```
yanfali: honestly we should fix this and point it at discord, maybe remove it for now

.travis.yml
```diff
@@ -12,6 +12,7 @@ script:
   - yarn run lint
   - yarn run test:unit
   - yarn run test:e2e
+  - yarn run build
```
yanfali: you actually don't need run when using yarn. It just does it if it doesn't match an internal command.

PR #461, opened 2019-08-06
Create coseyfannitutti_discipline_default.json

coseyfannitutti:


public/keymaps/coseyfannitutti_discipline_default.json
```diff
@@ -0,0 +1,22 @@
+{
+  "keyboard": "coseyfannitutti/discipline",
+  "keymap": "default",
```
noroadsleft: @coseyfannitutti, you can't actually use `default` as the value for `keymap`, because the API that supports Configurator can see the `keymap.c` files that are in qmk_firmware, even though Configurator itself can't do anything with them. In fact, the value for `keymap` cannot match *any* `keymap.c` that exists in qmk_firmware, for the same reason. The API tries to make a `keyboards/{keyboard}/keymaps/{keymap}/keymap.c` file when the compile job completes, but if one already exists the operation fails.

I append a partial commit hash to the `keymap` value to prevent this. (What you append actually doesn't matter so long as the corresponding file doesn't already exist in qmk_firmware.)

```suggestion
  "keymap": "default_430c370",
```

public/keymaps/coseyfannitutti_discipline_default.json
```diff
@@ -0,0 +1,22 @@
...
+  "commit": "",
```
noroadsleft: Similarly, I put the full hash of the last commit to modify the default keymap from qmk/qmk_firmware's master branch at the time I created/modified the keymap JSON. Configurator doesn't actually use the value, but I periodically check the keymaps to make sure the ones here match those in qmk_firmware as much as possible, and having the source hash makes that easier to do.

```suggestion
  "commit": "430c37024e8de42bb735f26386f2b13f042e3569",
```

public/keymaps/coseyfannitutti_discipline_default.json
```diff
@@ -0,0 +1,22 @@
...
+      "KC_TAB",  "KC_Q",    "KC_W",    "KC_E",    "KC_R",    "KC_T",    "KC_Y",    "KC_U",    "KC_I",    "KC_O",    "KC_P",    "KC_LBRC", "KC_RBRC", "KC_BSLS", "KC_DEL",
+      "KC_CAPS", "KC_A",    "KC_S",    "KC_D",    "KC_F",    "KC_G",    "KC_H",    "KC_J",    "KC_K",    "KC_L",    "KC_SCLN", "KC_QUOT", "KC_ENT",             "KC_PGUP",
+      "KC_LSFT",            "KC_Z",    "KC_X",    "KC_C",    "KC_V",    "KC_B",    "KC_N",    "KC_M",    "KC_COMM", "KC_DOT",  "KC_SLSH", "KC_RSFT", "KC_UP",   "KC_PGDN",
+      "KC_LCTL", "KC_LGUI", "KC_LALT",                                  "KC_SPC",                        "KC_RALT", "MO(1)",   "KC_RCTL", "KC_LEFT", "KC_DOWN", "KC_RGHT"
+    ],
+    [
+      "KC_GRV",  "KC_F1",   "KC_F2",   "KC_F3",   "KC_F4",   "KC_F5",   "KC_F6",   "KC_F7",   "KC_F8",   "KC_F9",   "KC_F10",  "KC_F11",  "KC_F12",  "KC_DEL",  "KC_PSCR",
+      "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_TRNS", "RESET",   "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_INS",  "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_PAUS", "KC_BSLS", "KC_TRNS",
```
noroadsleft: The `KC_BSLS` here is actually `KC_TRNS` on qmk_firmware, but as you're overriding `KC_BSLS` on Layer 0 with the same keycode on Layer 1, this isn't actually functionally different, and thus I'm not requiring it.

```suggestion
      "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_TRNS", "RESET",   "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_INS",  "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_PAUS", "KC_TRNS", "KC_TRNS",
```

coseyfannitutti:
@noroadsleft Okay, I was wondering exactly where those were being pulled from. Thanks for the explanation. I will pull the hash/partial hash from their respective keymap.c in the future.

PR #463, opened 2019-08-06
fix CNAME

zekth:
Fixing mixing CNAME in dist repository

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
   - yarn run test:unit
   - yarn run test:e2e
   - yarn run build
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,7 @@ script:
   - yarn run test:unit
   - yarn run test:e2e
   - yarn run build
+  - echo "config.qmk.fm" > dist/CNAME
```
zvecr: You should be able to use the deploy `fqdn` option <https://docs.travis-ci.com/user/deployment/pages/#further-configuration>

Issue #424, opened 2019-07-03
keep list of user's favorite keyboards in local storage and filter UI


yanfali:
Users would be able to pick which keyboards they prefer to see in their drop down, and toggle this on and off. Maybe have a :heart: icon which a user would click to add it to list and a :filter: icon to toggle the feature on and off.

Because all data is stored in the browser and never sent anywhere else this should not contravene GDPR because there is no PII (Personally Identifiable Information) beyond a simple list of keyboards in local storage, and possibly the layout.

zekth:
This may be done after the refactor involved by https://github.com/qmk/qmk_configurator/issues/415 ?

yanfali:
That's probably a good idea.

Issue #415, opened 2019-06-21
load from URL parameter


yanfali:
Now we have cloud loading from @zekth add URL parameter so people can send each other links pointing to keymaps

Issue #414, opened 2019-06-21
[chore] refactor startup ordering


yanfali:
Right now the logic for startup is spread out across a bunch of different components.
There are at least 2 aspects:

mounted - when the page initially loads - parsing URL parameters
watch - when someone actively changes a URL

components are doing this in multiple places.

TODO: move responsibility into App.vue/Main.vue and use properties on components to update behavior. Having this logic distributed is not maintainable.

zekth:
Totally agree with this. App statement as to be handled in the APP and not throught all the components.

PR #465, opened 2019-08-10
Refactor app loading / Favorite keyboard

zekth:
So i've tackled a refactoring of the loading of the state. Also added the favorite keyboard feature and started the bootstrap to load from URL.

Capture d’écran 2019-08-10 à 22 31 00

closes: https://github.com/qmk/qmk_configurator/issues/414
closes: https://github.com/qmk/qmk_configurator/issues/424

Will help for this:
closes: https://github.com/qmk/qmk_configurator/issues/415

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -9,7 +9,9 @@
         >
           <font-awesome-icon icon="star" size="lg" fixed-width />
         </a>
-        <label class="drop-label" id="drop-label-keyboard">{{ $t('message.keyboard.label') }}:</label>
+        <label class="drop-label" id="drop-label-keyboard"
+          >{{ $t('message.keyboard.label') }}:</label
+        >
         <v-select
           @search:focus="opened"
           @search:blur="blur"
@@ -25,7 +27,8 @@
           class="drop-label"
           :class="fontAdjustClasses"
           :title="$t('message.keymapName.label')"
-        >{{ $t('message.keymapName.label') }}:</label>
+          >{{ $t('message.keymapName.label') }}:</label
+        >
         <input
           id="keymap-name"
           type="text"
@@ -41,22 +44,29 @@
           id="load-default"
           :title="$t('message.loadDefault.title')"
           @click="loadDefault"
-        >{{ $t('message.loadDefault.label') }}</button>
+        >
+          {{ $t('message.loadDefault.label') }}
+        </button>
         <button
           id="compile"
           :title="$t('message.compile.title')"
           v-bind:disabled="compileDisabled"
           @click="compile"
-        >{{ $t('message.compile.label') }}</button>
+        >
+          {{ $t('message.compile.label') }}
+        </button>
       </div>
       <div class="topctrl-layouts">
-        <label class="drop-label" id="drop-label-version">{{ $t('message.layout.label') }}:</label>
+        <label class="drop-label" id="drop-label-version"
+          >{{ $t('message.layout.label') }}:</label
+        >
         <select id="layout" v-model="layout">
           <option
             v-for="(aLayout, layoutName) in layouts"
             :key="layoutName"
             v-bind:value="layoutName"
-          >{{ layoutName }}</option>
+            >{{ layoutName }}</option
+          >
         </select>
       </div>
     </div>
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -81,14 +81,11 @@ import first from 'lodash/first';
 import isUndefined from 'lodash/isUndefined';
 import isString from 'lodash/isString';
 
-import axios from 'axios';
 import { PREVIEW_LABEL } from '@/store/modules/constants';
 
 import {
   statusError,
   load_converted_keymap,
-  // render_layout,
-  // getExclusionList,
   compileLayout,
   disableOtherButtons
 } from '@/jquery';
@@ -204,7 +201,7 @@ export default {
       'startListening',
       'previewRequested'
     ]),
-    ...mapActions('app', ['fetchKeyboards']),
+    ...mapActions('app', ['fetchKeyboards', 'loadDefaultKeymap']),
     /**
      * loadDefault keymap. Attempts to load the keymap data from
      * a predefined known file path.
@@ -217,13 +214,11 @@ export default {
           return false;
         }
       }
-      let keyboardName = this.keyboard.replace(/\//g, '_');
-      let store = this.$store;
-      // TODO move this to store
-      axios
-        .get(`keymaps/${keyboardName}_default.json`)
-        .then(({ data, status }) => {
-          if (status === 200) {
+      const keyboardName = this.keyboard.replace(/\//g, '_');
+      const store = this.$store;
+      this.loadDefaultKeymap(keyboardName)
+        .then(data => {
+          if (data) {
             console.log(data);
             this.updateLayout(data.layout);
             let promise = new Promise(resolve =>
@@ -333,7 +328,7 @@ export default {
      * @return {undefined}
      */
     updateLayout(e) {
-      let newLayout = e.target ? e.target.value : e;
+      const newLayout = e.target ? e.target.value : e;
       this.setLayout(newLayout);
       this.$router.replace({ path: `/${this.keyboard}/${this.layout}` });
     },
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -293,7 +293,7 @@ export default {
         keyboardP !== '' &&
         keyboardP !== PREVIEW_LABEL
       ) {
-        // if someone loads a specific keyboard log it
+        // if someone loads a specific keyboard load it
         _keyboard = keyboardP;
         this.firstRun = false;
       }
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -2,7 +2,14 @@
   <div id="controller-top">
     <div class="topctrl">
       <div class="topctrl-keyboards">
-        <label class="drop-label">{{ $t('message.keyboard.label') }}:</label>
+        <a
+          id="favorite-keyboard"
+          v-on:click="favKeyboard"
+          v-bind:class="{ active: keyboard === favoriteKeyboard }"
+        >
+          <font-awesome-icon icon="star" size="lg" fixed-width />
+        </a>
+        <label class="drop-label" id="drop-label-keyboard">{{ $t('message.keyboard.label') }}:</label>
         <v-select
           @search:focus="opened"
           @search:blur="blur"
@@ -18,8 +25,7 @@
           class="drop-label"
           :class="fontAdjustClasses"
           :title="$t('message.keymapName.label')"
-          >{{ $t('message.keymapName.label') }}:</label
-        >
+        >{{ $t('message.keymapName.label') }}:</label>
         <input
           id="keymap-name"
           type="text"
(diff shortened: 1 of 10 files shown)


src/store/modules/app.js
```diff
@@ -161,6 +187,15 @@ const actions = {
...
     }
     commit('setDarkmode', darkStatus);
+  },
+  loadApplicationState({ dispatch }) {
+    dispatch('toggleDarkMode', true);
+    dispatch('loadFavoriteKeyboard');
+  },
+  loadFavoriteKeyboard({ state, commit }) {
+    if (state.favoriteKeyboard) {
+      commit('setKeyboard', state.favoriteKeyboard);
```
yanfali: we should probably check if this is a valid keyboard and if not, delete it

zekth: Valid keyboards means it is present in state.keyboards array?

yanfali: yeah, basically don't favorite something until after we have the full list of keyboards.

zekth:
ready to review @yanfali

zekth:
Regarding the e2e it's normal that it does not work because the API is not reachable by the travis CI.
Due to this change:
https://github.com/qmk/qmk_configurator/pull/465/files#diff-872d8e0270fe83b7bae281972a9c2a5dR4

The e2e are not working because the keyboards cannot be fetch. Before it was the same case but the UI was rendered empty. So i'm not really sure about the consistency of the e2e tests.

zekth:
Also might be cool to move to https://github.com/cypress-io/cypress

zekth:
As discussed with @yanfali , i've disabled the e2e tests

src/components/ControllerTop.vue
```diff
@@ -2,7 +2,18 @@
   <div id="controller-top">
     <div class="topctrl">
       <div class="topctrl-keyboards">
-        <label class="drop-label">{{ $t('message.keyboard.label') }}:</label>
+        <a
+          id="favorite-keyboard"
+          v-on:click="favKeyboard"
+          v-bind:class="{
+            active: keyboard === configuratorSettings.favoriteKeyboard
```
yanfali: let's use the same style as rest of code and turn this into a computed property

src/components/ControllerTop.vue
```diff
@@ -231,51 +246,60 @@ export default {
...
+      // if (this.$route.query.url) {
+      // }
+
+      let { keyboardP, layoutP } = this.$route.params;
+      if (
+        isString(keyboardP) &&
+        keyboardP !== '' &&
+        keyboardP !== PREVIEW_LABEL
+      ) {
+        // if someone loads a specific keyboard log it
```
yanfali:
```suggestion
        // if someone loads a specific keyboard load it
```

src/components/ControllerTop.vue
```diff
@@ -387,16 +418,27 @@ export default {
     };
   },
   mounted() {
-    this.fetchKeyboards().then(() => {
+    console.info('mounted start');
+    this.initializeKeyboards().then(() => {
```
yanfali: hmm, I wonder if we can't set this from props into the component from App.vue

src/components/SettingsPanel.vue
```diff
@@ -90,7 +90,7 @@ export default {
   components: { ToggleButton },
   computed: {
     ...mapState('keymap', ['continuousInput', 'displaySizes']),
-    ...mapState('app', ['tutorialEnabled', 'darkmodeEnabled'])
+    ...mapState('app', ['tutorialEnabled', 'configuratorSettings'])
```
yanfali: this would be a good use of a getter, since you could hide the implementation details

src/store/localStorage.js
```diff
@@ -0,0 +1,30 @@
+const CONSTS = {
+  configuratorSettings: 'configuratorSettings',
+  configuratorSettingsVersion: 1
+};
+
```
yanfali: I like how super generic this is, but I'm wondering if we should just have wrapped versions of this that we invoke from the users that remove some of the boiler plate

src/store/modules/app.js
```diff
@@ -73,6 +84,46 @@ const getters = {
 };
 
 const actions = {
+  checkValidKeymap(_, keymap) {
+    if (
```
yanfali: you can make this simpler, just return the entire thing without the true false

src/store/modules/app.js
```diff
@@ -73,6 +84,46 @@ const getters = {
...
+      isUndefined(keymap.keyboard) ||
+      isUndefined(keymap.keymap) ||
+      isUndefined(keymap.layout) ||
+      isUndefined(keymap.layers)
+    ) {
+      return false;
+    }
+    return true;
+  },
+  validateKeyboard({ state }, keyboard) {
```
yanfali: why is this an action?

zekth: getter is a property, action is a function. This does the validation, if the keyboard is in the availables keyboards.

src/store/modules/app.js
```diff
@@ -73,6 +84,46 @@ const getters = {
...
+      commit(
+        'setKeyboards',
+        r.data.filter(keeb => {
+          return isUndefined(exclude[keeb]);
+        })
+      );
+    }
+  },
+  loadDefaultKeymap(_, keyboardName) {
+    return axios.get(`keymaps/${keyboardName}_default.json`).then(r => {
```
yanfali: since you used async style above, I'm ok with that for this too

src/store/modules/app.js
```diff
@@ -73,6 +84,46 @@ const getters = {
...
+    const r = await axios.get(backend_keyboards_url);
+    if (r.status === 200) {
+      const exclude = getExclusionList();
+      commit(
+        'setKeyboards',
+        r.data.filter(keeb => {
+          return isUndefined(exclude[keeb]);
+        })
+      );
+    }
```
yanfali: you should return r

src/store/modules/app.js
zekth: Like a one liner? not really a fan personnaly as there is several conditions. I think it's more readable like this.

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
-  toggleDarkMode({ commit, state }, init) {
-    let darkStatus = state.darkmodeEnabled;
+  toggleDarkMode({ commit, state, dispatch }, init) {
+    let darkStatus = state.configuratorSettings.darkmodeEnabled;
     if (!init) {
       darkStatus = !darkStatus;
     }
     if (darkStatus) {
-      localStorageSet('darkmode', '1');
       document.getElementsByTagName('html')[0].dataset.theme = 'dark';
```
yanfali: I really don't like this, I'm wondering if we should just create a minimal component and move the dom manipulation in there controlled by a state value

src/store/localStorage.js
zekth: mmmm Yeah we might have a factory for the localstorage configuration objects. Which will parse it and make it usable with the current version of the configurator.

src/components/SettingsPanel.vue
zekth: As the property is directly accessed and not mutated getter is not mandatory IMO. But really depends on the way you want to get the infos from the config. We can do both yes.

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
     } else {
-      localStorageSet('darkmode', '0');
       document.getElementsByTagName('html')[0].dataset.theme = '';
     }
     commit('setDarkmode', darkStatus);
+    dispatch('saveConfiguratorSettings');
```
yanfali: probably need to return to make sure promises work correctly

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
+  },
+  setFavoriteKeyboard({ commit, dispatch }, keyboard) {
+    commit('setFavoriteKeyboard', keyboard);
+    dispatch('saveConfiguratorSettings');
```
yanfali: return?

src/components/ControllerTop.vue
zekth: MMM both way works in fact. Props may be more "component" way yes.

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
+  },
+  async loadApplicationState({ commit, dispatch }) {
+    console.log('loadApplicationState Start');
+    await dispatch('fetchKeyboards');
+    await dispatch('loadFavoriteKeyboard');
+    dispatch('toggleDarkMode', true);
+    console.log('loadApplicationState End');
+    commit('setAppInitialized', true);
```
yanfali: do we need a try catch and a return ?

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
+  },
+  async loadFavoriteKeyboard({ dispatch, state, commit }) {
+    if (state.configuratorSettings.favoriteKeyboard) {
+      if (
+        await dispatch(
+          'validateKeyboard',
```
yanfali: ok, I see why you need this to be an action.

src/store/modules/app.js
zekth: i don't agree with it, because when you call fetchKeyboards you don't want to deal with the HTTP response, just the data. Also this is just a void.

src/store/modules/app.js
zekth: In a further PR yes.

src/store/modules/app.js
zekth: Will do.

src/store/modules/app.js
```diff
@@ -148,20 +199,57 @@ const actions = {
...
+          state.configuratorSettings.favoriteKeyboard
+        )
+      ) {
+        console.info(
+          `setKeyboard ${state.configuratorSettings.favoriteKeyboard}`
+        );
+        commit('setKeyboard', state.configuratorSettings.favoriteKeyboard);
+      } else {
+        console.info('erase wrong keyboard');
```
yanfali:
```suggestion
        console.info('Invalid keyboard favorited. Removing setting from local storage');
```

src/store/modules/app.js
```diff
@@ -267,6 +358,9 @@ const mutations = {
     }
     return {};
   },
+  setAppInitialized(state, status) {
+    state.appInitialized = status;
```
yanfali: in the normal case we should never need to toggle this. Maybe just remove status and set it to true

src/store/modules/app.js
```diff
@@ -316,7 +410,7 @@ const mutations = {
     state.tutorialEnabled = !state.tutorialEnabled;
   },
   setDarkmode(state, enabled) {
-    state.darkmodeEnabled = enabled;
+    state.configuratorSettings.darkmodeEnabled = enabled;
```
yanfali: configurator settings is starting to smell, I wonder if it just needs to be it's own module

src/store/modules/app.js
zekth: try catch would be overkill here as it's already catch in nested functions. What do you want to return? It's a void.

src/store/modules/app.js
zekth: no. This is due to this: https://github.com/qmk/qmk_configurator/pull/465/files/f6ff2859823fb65a818063a2b232bf4bf25dabba#diff-872d8e0270fe83b7bae281972a9c2a5dR4

but if we mode the keyboards to props, this state is no more revelant

zekth:
I made the changes we discussed. All other changes may move in other PRs unless this PR will be too big.

src/store/modules/app.js
yanfali: I still think returning the dispatch would be better from a chaining point of view, but this is fine.

src/store/modules/app.js
```diff
@@ -73,6 +89,41 @@ const getters = {
 };
 
 const actions = {
+  checkValidKeymap(_, keymap) {
+    if (
```
yanfali: Seems about the same, with less boilerplate.

```suggestion
    return isUndefined(keymap.keyboard) || isUndefined(keymap.keymap)  || isUndefined(keymap.layout) || isUndefined(keymap.layers);
```

src/store/modules/app.js
yanfali: probably not explaining myself well, but when do we ever set `appInitialized = false`?

src/store/modules/app.js
```diff
@@ -73,6 +89,33 @@ const getters = {
 };
 
 const actions = {
+  checkValidKeymap(_, keymap) {
```
yanfali: TODO figure out what the _ is for

zekth: the object passed with `{state, commit ... }` . So as a convention, because we are not using it, i've put `_`

PR #467, opened 2019-08-14
Initial implementation of code for electron App

e11i0t23:
submitted to allow for a wider testing without having to host this on local machine for each test OS for flashing configuration. Also in preperation for beta release for wider testing of https://github.com/e11i0t23/qmk_configurator_electron

The Code added to main.js is required to detect weather the site is being run inside electron app or inside a web browser. In the first case we then share a simple function to allow for adding text to terminal with the app

When run in electron the download button is replaced by two alternative buttons, one for flashing a file compiled by config as well as an alternative for flashing a user pre-compiled this helps with testing the app and its multitude of flashers without having to recompile firmware but has been left in as is a useable feature.

### DIFF
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -98,19 +98,21 @@
         class="fixed-size"
         id="fwFile"
         @click="autoFlashFirmware"
-        title="Automaticly Flash compiled Firmware to MCU"
+        :title="$t('message.flashFirmware.title')"
         v-bind:disabled="disableDownloadBinary"
       >
-        <font-awesome-icon icon="download" size="lg" fixed-width />Auto-Flash
+        <font-awesome-icon icon="download" size="lg" fixed-width />
+        {{ $t('message.flashFirmware.label') }}
       </button>
       <button
         class="fixed-size"
         id="fwFile"
         @click="flashFirmware"
-        title="Flash User Selected file to MCU"
+        :title="$t('message.flashFile.title')"
         v-bind:disabled="disableFlashFile"
       >
-        <font-awesome-icon icon="download" size="lg" fixed-width />Custom-Flash
+        <font-awesome-icon icon="download" size="lg" fixed-width />
+        {{ $t('message.flashFile.label') }}
       </button>
     </div>
     <div v-else class="botctrl-1-2">
@@ -120,7 +122,8 @@
         :title="$t('message.downloadFirmware.title')"
         v-bind:disabled="disableDownloadBinary"
       >
-        <font-awesome-icon icon="download" size="lg" fixed-width />{{ $t('message.downloadFirmware.label') }}
+        <font-awesome-icon icon="download" size="lg" fixed-width />
+        {{ $t('message.downloadFirmware.label') }}
       </button>
     </div>
     <div v-if="downloadElementEnabled">
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -166,7 +166,7 @@ export default {
       'author',
       'notes'
     ]),
-    ...mapGetters(['exportKeymapName']),
+    ...mapGetters(['exportKeymapName', 'firmwareFile']),
     disableDownloadKeymap() {
       return !this.enableDownloads && this.keymapSourceURL !== '';
     },
@@ -286,8 +286,8 @@ export default {
       window.Bridge.autoFlash = true;
       window.Bridge.flashURL(
         first(this.firmwareBinaryURL),
-        this.$store.getters['app/keyboard'],
-        this.$store.getters['app/firmwareFile']
+        this.keyboard,
+        this.firmwareFile
       );
     },
     downloadSource() {
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -94,26 +94,8 @@
       />
     </div>
     <div v-if="electron" class="botctrl-1-2">
-      <button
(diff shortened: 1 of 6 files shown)


src/components/ControllerBottom.vue
```diff
@@ -243,6 +278,18 @@ export default {
...
     },
+    flashFirmware() {
+      window.Bridge.autoFlash = false;
+      window.Bridge.flashFile();
+    },
+    autoFlashFirmware() {
+      window.Bridge.autoFlash = true;
+      window.Bridge.flashURL(
+        first(this.firmwareBinaryURL),
+        this.$store.getters['app/keyboard'],
```
zekth: use `mapGetters` like here : https://github.com/qmk/qmk_configurator/blob/801795e4b38ec1f3be1986ac648dc479ab7b8fd3/src/components/ControllerBottom.vue#L169

src/main.js
```diff
@@ -46,6 +46,23 @@ import {
 } from '@fortawesome/free-brands-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ga from './ga';
+// Electron specific Code
```
zekth: Is electron specific code is meant to be extended? Is it revelant to wrap it into a `util` module?

e11i0t23: I currently have no plan to extend it as the electron app solely needs to be able to write to the status screen other than that it handles everything else. But can move it to a util module if that is preferred

zekth: i think it could be nice, therefore if needed it would not require any refactor. Don't you think?

e11i0t23: I will implement this change later when i fix the other two issues then for future eas incase i need to implement anything else

src/components/ControllerBottom.vue
e11i0t23: implemented now, keyboard already existed as a state which was useable but added firmwareFile

src/main.js
e11i0t23: I have left in the if statement in main to check and then if in electron it calls the init function from electron.js allowing code to be added to this in the future and no changes to be needed to main js for these

src/components/ControllerBottom.vue
```diff
@@ -157,6 +179,19 @@ export default {
...
+      return !window.Bridge.enableFlashing;
+    },
+    disableFlashSource() {
+      return (
+        !window.Bridge.enableFlashing ||
+        isUndefined(this.firmwareBinaryURL) ||
+        this.firmwareBinaryURL === ''
+      );
+    },
+    electron: function() {
```
yanfali: why is this an anon function?

src/main.js
```diff
@@ -46,6 +46,17 @@ import {
 } from '@fortawesome/free-brands-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ga from './ga';
+import electron from './electron';
+// Find out if we are are running inside electon
+window.electron = false;
```
yanfali: just set a value on the app store, try to avoid touching window name space.

src/store/modules/app.js
```diff
@@ -51,6 +51,7 @@ const steno_keyboards = ['gergo', 'georgi'];
 const getters = {
   firmwareFile: state => state.firmwareFile,
   filter: state => state.filter,
+  keyboard: state => state.keyboard,
```
yanfali: don't do this please. Use mapState and just get it directly. Getters should only be used when the value needs to be changed in flight, all these getters probably need to go away.

yanfali:
@e11i0t23 I love how little code this is, good job. I do have a few issues with the way we're interacting with bridge though. My preference would be to pull in a new component for the new controls and just pull it in that way and have the  handlers inside that new component. Let me know if you want to talk about how to do this. It should be fairly trivial.

src/electron.js
```diff
@@ -0,0 +1,14 @@
+import store from './store';
+
+export default {
+  init() {
+    store.commit('app/enableElectron')
+
+    //We use the Bridge as a way to share functions between electron and vue
+    window.Bridge.statusAppend = txt => {
+      txt = '\n' + txt;
```
yanfali: prefer backticks style over assembling strings, it's cleaner and less error prone
```
  txt = `\n${txt}`;
```

src/main.js
```diff
@@ -46,6 +46,14 @@ import {
 } from '@fortawesome/free-brands-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ga from './ga';
+import electron from './electron';
+// Find out if we are are running inside electon
+if (
+  typeof navigator === 'object' &&
```
yanfali: I prefer to use the lodash helpers, they catch more corner cases. import isObject and isString/

https://lodash.com/docs/4.17.15#isObject
https://lodash.com/docs/4.17.15#isString

Use `includes` as it's convey's intent better than `indexOf`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

src/main.js
```diff
@@ -46,6 +46,15 @@ import {
 } from '@fortawesome/free-brands-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ga from './ga';
+import { isObject, isString } from 'lodash';
```
yanfali:
```suggestion
import isObject from 'lodash/isObject';
import isString from 'lodash/isString';
```

Issue #471, opened 2019-08-17
Refactor UI tests


zekth:
Like mentionned in the https://github.com/qmk/qmk_configurator/pull/465 

We need to refactor UI testing and do better testing, because ATM it's just trying to check if the DOM is present but nothing is really tested.

Also instead of using selenium i suggest to use Cypress:
https://github.com/cypress-io/cypress

I'll make a proposal for it

PR #472, opened 2019-08-17
replace selenium by cypress

zekth:
closes: https://github.com/qmk/qmk_configurator/issues/471
Easier to deploy, to develop, to use.
anyway it's better.

Et voilà

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "chromedriver": "^74.0.0",
     "cypress": "^3.4.1",
     "eslint": "^5.8.0",
+    "eslint-plugin-cypress": "^2.6.1",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-vue": "^5.0.0-0",
     "growl": "^1.10.0",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
     "test:e2e": "vue-cli-service test:e2e",
+    "test:cypress": "cypress run",
+    "test:cypress:ci": "start-server-and-test serve http://localhost:8080 test:cypress",
     "test:unit": "vue-cli-service test:unit"
   },
   "dependencies": {
@@ -40,6 +42,7 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.0.0",
     "chromedriver": "^74.0.0",
+    "cypress": "^3.4.1",
     "eslint": "^5.8.0",
     "eslint-plugin-prettier": "^3.1.0",
     "eslint-plugin-vue": "^5.0.0-0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -7,6 +7,8 @@
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
     "test:e2e": "vue-cli-service test:e2e",
```
yanfali: remove this?

package.json
```diff
@@ -7,6 +7,8 @@
...
+    "test:cypress": "cypress run",
```
yanfali: what does this task do? It doesn't run standalone

tests/integration/start.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      '30wer'
+    ]);
+  });
+  it('Starts the app Check the dom', function() {
+    cy.visit('/');
+    cy.get('#controller-top').should('be.visible');
+    cy.get('#visual-keymap').should('be.visible');
+    cy.get('#status').should('be.visible');
+    cy.get('#keycodes-section').should('be.visible');
+    cy.get('.openSettings').click();
```
yanfali: split this into it's own test.

tests/integration/start.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    cy.get('.settings-panel', { timeout: 5000 }).should('be.visible');
+    cy.get('.slideout-panel-bg').click();
+    cy.get('#testkeys').click();
```
yanfali: same, should be it's own test. You're changing the route

package.json
zekth: It could but easier to have everything starting directly from yarn

yanfali: ok, I think I see what you mean now. Maybe put a comment on this.

src/electron.js
```diff
@@ -2,12 +2,14 @@ import store from './store';
 
 export default {
   init() {
-    store.commit('app/enableElectron')
+    store.commit('app/enableElectron');
```
yanfali: should we move this inside the guard?

tests/integration/start.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      '1upkeyboards/1up60hte',
+      '1upkeyboards/1up60rgb',
+      '1upkeyboards/super16',
+      '1upkeyboards/sweet16',
+      '2_milk',
+      '30wer'
+    ]);
+    cy.visit('/');
+  });
+  it('Top Controller should be instanciated and visible', function() {
```
yanfali:
```suggestion
  it('Top Controller should be instantiated and visible', function() {
```

yanfali:
@e11i0t23 heads up, we're moving the electron code behind a guard to check for the presence of window.Bridge. Let us know if this causes you any issues. Thanks.

PR #473, opened 2019-08-19
Colorway: add dsa milkshake

zekth:
With approval from Biip
Screenshot 2019-08-19 at 12 05 15

### DIFF
--- a/src/components/colorways/index.js
+++ b/src/components/colorways/index.js
@@ -1,5 +1,5 @@
 import dsa_galaxy_class from './dsa/galaxy_class';
-import milkshake from './dsa/milkshake';
+import dsa_milkshake from './dsa/milkshake';
 import gmk from './gmk';
 import sa from './sa';
 import alphaCodes from './alphaCodes';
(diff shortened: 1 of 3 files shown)


zekth:
@yanfali its really weird because i just use 'yarn lint'

noroadsleft:
Differing lint settings, perhaps? I always do my linting by hand. :man_shrugging: 

This is more a stylistic complaint, but I don't like the way pure white keys look in light mode. Something like `background: #fdfbf7` for `.dsa-milkshake-key` and `.dsa-milkshake-mod` might look a bit nicer (though I'm not sure how true-to-life that would be as I don't own this set). Just something a bit off-white to bring out the keys against the light mode background.

Also I realized while testing this that there's a bug of sorts with key legends in dark mode versus light mode. Should be fixed, but out-of-scope here.

zekth:
I've asked biip and the white is just pure white #FFF so...

About linting it uses the lint config defined in the project because i only use `yarn lint`. But on this project it's totally weird and i don't really like the abstract way of the vue-clie to use prettier. I'd better use raw prettier command.

noroadsleft:
K, maybe a different color for `#visual-keymap` would help overall, which would be for another PR.

I'll defer to @yanfali as for the lint/prettier implementation.

src/components/colorways/index.js
```diff
@@ -1,4 +1,5 @@
 import dsa_galaxy_class from './dsa/galaxy_class';
+import milkshake from './dsa/milkshake';
```
yanfali: use the same naming convention. There's going to be a Kat Milkshake soon.

PR #480, opened 2019-08-25
Linting errors

yanfali:
- move a standalone function to jquery

@zekth  I think you line lengths are getting overriden in your local environment. I would prefer to keep short lines as it's more readable. Also I prefer standalone functions just live in their own file for now, not in an action. Action isn't an appropriate use for that since it never interacts with model.

 - break app.js into smaller pieces as it's getting unwieldy
 - make loadDefaultKeymap keep details of implementation within store
 - remove dead code from component


yanfali:
Removed some dead code and move implementation detail from UI component for default keymaps into store.

zekth:
Not sure about the linting errors IMO, because we use the same env and i have no global settings for my linting. As we already discussed, i think the problem comes from how the vue-cli-service handles it. It would be interesting to try raw prettier command on each of our envs.

src/components/ControllerTop.vue
```diff
@@ -372,34 +375,6 @@ export default {
...
-    },
-    logLoadDefaultSuccess(keyboardName) {
-      this.$ga.event({
-        eventCategory: 'apicall',
-        eventAction: 'loadDefaultSuccess',
-        eventLabel: keyboardName
-      });
-    },
-    logLoadDefaultFail(keyboardName) {
-      this.$ga.event({
```
zekth: Deleting this, we might delete ga.js?

yanfali: Eventually. When editing a file if you see it, we can start removing it. Don't want to blow up the pr.

PR #481, opened 2019-08-25
Prettier cmd

zekth:
Following: https://github.com/qmk/qmk_configurator/pull/480
Here is the addition of raw prettier commands because we are experiencing different behaviours using the vue-cli commands.

Here 2 commits, one pushing the commands, another pushing the modifications using the `prettier --write`.

Want to run it througth travis to see its behaviour


zekth:
So seems to work properly.
@yanfali your thoughts? could you test on your env?

Also regarding the ci:
https://travis-ci.org/qmk/qmk_configurator/builds/576405648#L453

This triggers me, because it throws errors but the CI is not blocked. Which is wrong no? Or maybe we could just disable those rules.

zekth:
The console log rule can be deleted IMO

zekth:
> > The console log rule can be deleted IMO
> 
> I think it's just a matter of setting it off. Right now it's using development/production to decide the rule. If you delete it, it defaults to on

was meaning setting it to:
```
rules: {
        'no-console': 'off',
    },
```

zekth:
Done

yanfali:
@zekth can you rebase this one and let's try and get it merged.

zekth:
@yanfali done

PR #482, opened 2019-08-25
fixing chatter detection

zekth:
After few tests i stumbleupon a weird behavior on `LT()` commands. This showed me that our chatter detection algorithm was wrong.

Now it compares the time between the last keyup and keydown event, which is the same behaviour as Switch hitter.

I have compared it with both of my boards and it works.

### DIFF
--- a/src/components/VisualTesterKeymap.vue
+++ b/src/components/VisualTesterKeymap.vue
@@ -26,21 +26,21 @@
       <h3 class="info-title">{{ $t('message.tester.keycodeStatus.label') }}</h3>
       <div class="letter-display">
         <div class="letter-key">
-          <label class="key-label">
-            {{ $t('message.tester.letters.key.label') }}
-          </label>
+          <label class="key-label">{{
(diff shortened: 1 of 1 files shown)


noroadsleft:
I'm not sure if I'm misunderstanding the issue or not testing this properly, but I'm not seeing any difference in results between current master and this branch. (I use an `LT(_FN, KC_CAPS)` key.)

zekth:
@noroadsleft first the trigger of the chatter was between 2 events keyup which is wrong, it's between the last keyup and the next keydown.

Also regarding the LT() issues i was having problem on this keymap: 
https://github.com/qmk/qmk_firmware/blob/master/keyboards/ut472/keymaps/default/keymap.c

with: `LT(3, KC_TAB)` and `KC_SFTENT`

zekth:
After few debugging and discussion. Closing it. Will document in https://github.com/qmk/qmk_firmware/tree/master/docs

PR #485, opened 2019-08-31
Update default keymaps for Vitamins Included

noroadsleft:
Revisions 1 and 2. Doing a teeny, tiny favor for @Duckle29.


Duckle29:
Thanks :D

PR #489, opened 2019-09-06
updated the french translation

gab-a:
*Added and translated lines missing in index.js

*Created and translated print.js and tester.js

Sorry @yanfali I failed to test my changes with node. I am not very good with all this stuff yet , hopefully I didn't break anything.

I haven't got any difficulties with the translations, I think they work. Except maybe for the part about the chatter threshold I couldn't find something I liked and very natural. Maybe @zekth would have a word to say

I think that's it,

have a nice weekend everyone

### DIFF
--- a/src/i18n/fr/index.js
+++ b/src/i18n/fr/index.js
@@ -119,9 +119,9 @@ export default {
           help: 'Montre la taille des touches (une touche classique = 1 u)'
         },
         toggleTutorial: {
-          label: 'Vidéo Guide',
-          title: 'Vidéo guide de MechMerlin sur le configurateur (en Anglais)',
-           help: 'Vidéo guide de MechMerlin (en Anglais)'
+          label: 'Guide Vidéo',
+          title: 'Guide vidéo de MechMerlin sur le configurateur (en Anglais)',
+           help: 'Guide vidéo de MechMerlin (en Anglais)'
         },
         darkmode: {
           label: 'Mode Sombre',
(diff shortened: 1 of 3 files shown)


src/i18n/fr/index.js
```diff
@@ -94,6 +118,15 @@ export default {
           title: 'ctrl + alt + u',
           help: 'Montre la taille des touches (une touche classique = 1 u)'
         },
+        toggleTutorial: {
+          label: 'Vidéo Guide',
```
zekth: Would be `Guide vidéo` no?

src/i18n/fr/tester.js
```diff
@@ -0,0 +1,35 @@
...
+      label: 'touche'
+    }
+  },
+  docs: {
+    paragraph:
+      'Note: La touche et le code détecté.es peuvent varier selon votre localisation. Docs'
+  },
+  chatter: {
+    label: 'Seuil de détection d\'activations répétées (ms)',
+    detectedAlert: 'Activations répétées détectées'
```
zekth: Mention of `Chatter` would be a good addition like:
`Activations répétées détectées (Chatter)`
Because it's the universal way to call it.

src/i18n/fr/tester.js
```diff
@@ -0,0 +1,35 @@
...
+    code: {
+      label: 'code'
+    },
+    key: {
+      label: 'touche'
+    }
+  },
+  docs: {
+    paragraph:
+      'Note: La touche et le code détecté.es peuvent varier selon votre localisation. Docs'
```
zekth: Not sure about the inclusive writing here, looks like a typo. Also the good dot character is this one `·` no the classic `.`.

src/i18n/fr/tester.js
gab-a: I agree, good idea thank you

PR #490, opened 2019-09-09
Add a more helpful when redis is failing

yanfali:
- use the job count to figure out when we are having service issues and
 display error information


noroadsleft:
Also, it'd probably be good if https://github.com/qmk/qmk_configurator/blob/7c5cb97a2801f04d1a83040a9af5eeae780cc165/src/components/StatusBar.vue#L60 actually linked to QMK Discord (#configurator channel if possible).

PR #492, opened 2019-09-11
Add One-Shot Mods

noroadsleft:
Adds One-Shot Mods to the Quantum tab.

Prompted by Belemrys#4678 on QMK Discord.

### DIFF
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -120,12 +120,6 @@ function processOneShotMods(keycode) {
   keycode = `OSM(${mods})`;
 
   const metadata = store.getters['keycodes/lookupKeycode'](keycode);
-  if (metadata === undefined) {
-    // it's not valid - return an ANY key
-    stats.any += 1;
-    return newAnyKey(keycode);
-  }
-  // it's valid - return a keycode
   return newKey(metadata, keycode);
 }
 
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -40,7 +40,14 @@ function init() {
   store.commit('app/setKeypressListener', () => keypressListener);
 }
 
-// returns true if bit `value` of `test` is true
+/* hasBitsSet
+ *
+ * arguments:
+ *   test (number)
+ *   value (number)
+ *
+ * returns true if bit `value` of `test` is true, false otherwise
+ */
 function hasBitsSet(test, value) {
   return (test & (1 << value)) === Math.pow(2, value);
 }
--- a/src/jquery.js
+++ b/src/jquery.js
@@ -46,42 +46,90 @@ function hasBitsSet(test, value) {
 }
 
 // check One-Shot Mod keycodes
-function checkOneShotMods(keycode) {
-    let maincode = keycode.split('(')[0];
-    let internal = keycode.split('(')[1];
-    internal = internal.split(')')[0];
+function processOneShotMods(keycode) {
+  let maincode = keycode.split('(')[0];
+  let internal = keycode.split('(')[1];
+  internal = internal.split(')')[0];
 
-    // tokenizers
-    let mods = internal.split('|');
-    mods = mods.map((amod) => {
-        return amod.trim();
-    });
+  // tokenizers
+  let mods = internal.split('|');
+  mods = mods.map((amod) => {
+    return amod.trim();
+  });
 
-    // parser
-    mods = mods.map((amod) => {
-        // MOD_LCTL = 0x0001,
-        // MOD_LSFT = 0x0010,
-        // MOD_LALT = 0x0100,
-        // MOD_LGUI = 0x1000,
-        switch (amod) {
-        case 'MOD_LCTL':
-        case 'MOD_RCTL':
-            return 0b0001;
-        case 'MOD_LSFT':
-        case 'MOD_RSFT':
-            return 0b0010;
-        case 'MOD_LALT':
-        case 'MOD_RALT':
-            return 0b0100;
-        case 'MOD_LGUI':
-        case 'MOD_RGUI':
-            return 0b1000;
-        case 'MOD_MEH':
-            return 0b0111;
-        case 'MOD_HYPR':
-            return 0b1111;
-        }
-    });
+  // parser
+  mods = mods.map((amod) => {
+    // MOD_LCTL = 0x0001,
+    // MOD_LSFT = 0x0010,
+    // MOD_LALT = 0x0100,
+    // MOD_LGUI = 0x1000,
+    switch (amod) {
+      case 'MOD_LCTL':
+      case 'MOD_RCTL':
+        return 0b0001;
+      case 'MOD_LSFT':
+      case 'MOD_RSFT':
+        return 0b0010;
+      case 'MOD_LALT':
+      case 'MOD_RALT':
+        return 0b0100;
+      case 'MOD_LGUI':
+      case 'MOD_RGUI':
+        return 0b1000;
+      case 'MOD_MEH':
+        return 0b0111;
+      case 'MOD_HYPR':
+        return 0b1111;
+    }
+  });
+
+  // code generator
+  mods = mods.reduce((acc, amod) => {
+    acc |= amod;
+    return acc;
+  });
+
+  let cmods = [];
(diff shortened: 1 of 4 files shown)


src/store/modules/keycodes/quantum.js
```diff
@@ -8,6 +8,34 @@ function makeLT(layer) {
...
+    'RSft': ['RSft' , 'Right Shift'],
+    'RCtl': ['RCtl' , 'Right Control'],
+    'RAlt': ['RAlt' , 'Right Alt'],
+    'RGUI': ['RGUI' , 'Right GUI'],
+    'Hypr': ['Hyper', 'Left Control, Shift, Alt and GUI'],
+    'Meh' : ['Meh'  , 'Left Control, Shift, and Alt']
+}
+function makeOSM(mod) {
+  // visual key legend
+  let text = osmLookup[mod][0];
```
yanfali: this will break if mod is undefined

src/store/modules/keycodes/quantum.js
```diff
@@ -8,6 +8,34 @@ function makeLT(layer) {
...
+  // actual keycode argument
+  let mod_t = "MOD_"+ mod.toUpperCase();
+  // help text
+  let help = osmLookup[mod][1];
```
yanfali: ditto

src/jquery.js
```diff
@@ -264,6 +264,83 @@ function parseKeycode(keycode, stats) {
     let internal = splitcode[1];
     internal = internal.split(')')[0];
 
+    // check for an OSM keycode
```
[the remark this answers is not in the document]
yanfali: throw me a branch to look at and I can see what's up

src/store/modules/keycodes/quantum.js
noroadsleft: I never actually figured out what to do about this.

yanfali: so typically you guard against mod not being a valid value. checking that it resolves first, so that the 2nd index lookup doesn't fail. It's not a huge deal because this is used internally. Just be aware it could crash if someone feeds it bad data in an update.

src/store/modules/keycodes/index.js
```diff
@@ -1,11 +1,11 @@
-import quantum from './quantum';
 import ansi from './ansi';
 import iso_jis from './iso-jis';
+import quantum from './quantum';
 import settings from './kb-settings';
 import media from './app-media-mouse';
 import steno from './steno';
 const state = {
-  keycodes: [...quantum, ...ansi, ...iso_jis, ...settings, ...media]
+  keycodes: [...ansi, ...iso_jis, ...quantum, ...settings, ...media]
```
yanfali: I feel like we're violating Don't repeat yourself, should probably just assign this once else where as a const and re-use it.

src/jquery.js
```diff
@@ -40,6 +40,89 @@ function init() {
   store.commit('app/setKeypressListener', () => keypressListener);
 }
 
+// returns true if bit `value` of `test` is true
+function hasBitsSet(test, value) {
+  return (test & (1 << value)) === Math.pow(2, value);
```
yanfali: why not use 1 << on right side too?

src/store/modules/keycodes/quantum.js
```diff
@@ -8,6 +8,46 @@ function makeLT(layer) {
...
+  'MOD_LALT|MOD_LGUI': ['AG', 'Alt and GUI'],
+  'MOD_LCTL|MOD_LSFT|MOD_LGUI': ['CSG', 'Control, Shift, and GUI'],
+  'MOD_LCTL|MOD_LALT|MOD_LGUI': ['CAG', 'Control, Alt, and GUI'],
+  'MOD_LSFT|MOD_LALT|MOD_LGUI': ['SAG', 'Shift, Alt, and GUI']
+};
+function makeOSM(mod, width = 1000) {
+  //if ( osmLookup[mod] ) {
+  // visual key legend
+  let text = osmLookup[mod][0];
+  /*  // actual keycode argument
```
yanfali: delete all this commented out code.

src/store/modules/keycodes/quantum.js
```diff
@@ -8,6 +8,46 @@ function makeLT(layer) {
     title: `kc on tap, switch to layer ${layer} while held`
   };
 }
+// make a One-Shot Mod Keycode Definition
+const osmLookup = {
+  MOD_LSFT: ['LSft', 'Left Shift'],
```
yanfali: does this need to be an array anymore? you're aren't using the 2nd value

src/store/modules/keycodes/quantum.js
```diff
@@ -241,6 +281,34 @@ export default [
...
+  makeOSM('MOD_LSFT|MOD_LGUI'),
+  makeOSM('MOD_LALT|MOD_LGUI'),
+  { width: 250 },
+  makeOSM('MOD_LCTL|MOD_LSFT|MOD_LGUI'),
+  makeOSM('MOD_LCTL|MOD_LALT|MOD_LGUI'),
+  makeOSM('MOD_LSFT|MOD_LALT|MOD_LGUI'),
+  { width: 250 },
+  makeOSM('MOD_MEH'),
+  makeOSM('MOD_HYPR'),
+  //makeOSM('MOD_TEST'),
```
yanfali: delete this please

src/jquery.js
```diff
@@ -40,6 +40,89 @@ function init() {
...
+}
+
+// check One-Shot Mod keycodes
```
yanfali: expand this comment please. basically explain you're normalizing out of order osm mods that were entered by hand potentially

yanfali:
I feel like this could use some unit tests since it's got some subtle behaviors like MEH and HYPR. Fix up the last couple of issues and we'll merge it. Thanks

src/store/modules/keycodes/quantum.js
noroadsleft: Yeah, I get what the problem is. I just couldn't think of a way to check that the values are valid.

src/store/modules/keycodes/index.js
noroadsleft: This is just restoring what's already on qmk:master. I changed it for development but accidentally committed the change.

src/jquery.js
noroadsleft: I've just always used `Math.pow()` this way, if that's what you're asking.

src/store/modules/keycodes/quantum.js
noroadsleft: It's on the agenda.

src/store/modules/keycodes/quantum.js
noroadsleft: I'm not using the second value because I never figured out how to check that `mod` is valid. I still intend to use this data.

src/jquery.js
yanfali: it's just weird. 1 << is orders of magnitude faster than Math.pow because it's an intrinsic CPU instruction, whereas Math.pow is a generalized power function. I would prefer you to use one or the other, but not both.

yanfali: I've also been informed that there is now a ** operation so you know longer need to call Math

https://mariusschulz.com/blog/the-exponentiation-operator-in-javascript

yanfali:
@noroadsleft is this good to go?

noroadsleft:
> @noroadsleft is this good to go?

I thought it was, but I just found a bug... or a case I didn't consider... or something.

Basically, the UI has both Left and Right OSM keycodes for Ctrl, Shift, Alt and GUI, and correctly assigns the Left or Right Mod based on what the user places. But when a keymap.json with OSM keycodes is imported, any OSM keycodes that were Right hand are assigned as Left hand. It's been so long since I've been under the hood on this code that I need to refresh how it works.

noroadsleft:
@yanfali Fixed the bug. Will take suggestions on whether 9ebc3504dead9e3ea70b288ee954029deb17b524 is the best/cleanest way to go about this.

noroadsleft:
@yanfali 504d4fdc55d57ced5bf4193fe6c58a476c85c227 ?

PR #493, opened 2019-09-12
Replace ctrl-esc with ctrl-bspc, add brackets to ctrl

jceb:


jceb:
Okay, thank you for the info. I updated the keymap accordingly.

Issue #501, opened 2019-09-23
Clear Layer function breaks layout key count and key assignment if the function is run from Layer 0


noroadsleft:
## Steps to Reproduce

1. Load any keyboard in QMK Configurator
2. While Layer 0 is selected, click the Clear Layer button and confirm when the dialog prompt appears.

The key count that appears at the lower right of the keymap will read `0 Keys`.

Any of the following restore the key count:

- Select the last key of the layout while on Layer 0 and assign a keycode to it (assuming you know which key is the last one)
- Importing a keymap (manually, or switch to Keyboard Tester or Print Keymap and back)
- Change the currently selected layout (e.g. ANSI to ISO), if there is another option

yanfali:
Good find. This is an internal implementation detail leaking into the UI. I think the easiest way to fix it is use the layout macro count.

sandyjoshi:
@yanfali  is this issue fixed?

noroadsleft:
@sandyjoshi Yes, this issue was fixed in the pull request referenced above.

PR #502, opened 2019-09-23
Fix keycount when clearing base layer

yanfali:
Fix for issue #501 

use the layout count rather than the keymap as we clear the base
layer and rely on the visual keymap to not send us incorrect
indices.  It may contain less than the entire keymap codes but
this is an internal detail that shouldn't be exposed to the UI.

The currently selected layout should always contain the correct number
of keys.

 - use a getter as this way it's cached until the layout or keyboard
 changes.

committed: #       modified:   src/components/Main.vue #       modified:
src/store/modules/app/getters.js # # Untracked files: #       CNAME

### DIFF
--- a/src/components/Main.vue
+++ b/src/components/Main.vue
@@ -42,7 +42,7 @@
         </p>
         <visualKeymap :profile="false" />
         <span class="keymap--count"
-          ><span class="keymap--counter">{{ size() }}</span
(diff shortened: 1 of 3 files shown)


yanfali:
Lol. I didn't test fast input. that's an awesome bug.

PR #518, opened 2019-10-04
add env variable handling for API url

### LINKED ISSUE #413: Allow environment variable override for backend_baseurl
it's OK that we hardcode backend_baseurl as a constant, but we need a way to override it during development without having to manually edit the config file.

This will be particularly useful for the docker-compose web stack at https://github.com/qmk/qmk_web_stack

zekth:
fix #413 
Based on the vue cli constaints : https://cli.vuejs.org/guide/mode-and-env.html#modes

.env file is loaded and can be modified for a local development environment.
.env.production is used when we use `yarn build`

Also if the runtime doesn't find the variable, it throw exception at startup.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+VUE_APP_API_URL=https://api.qmk.fm
--- a/.env.production
+++ b/.env.production
@@ -1,0 +1,1 @@
+VUE_APP_API_URL=https://api.qmk.fm
--- a/src/store/modules/constants.js
+++ b/src/store/modules/constants.js
@@ -1,6 +1,9 @@
 import template from 'lodash/template';
 
-export const backend_baseurl = 'https://api.qmk.fm';
+export const backend_baseurl = process.env.VUE_APP_API_URL;
+if (!backend_baseurl) {
+  throw 'Backend URL has not been specified';
+}
 export const backend_keyboards_url = `${backend_baseurl}/v1/keyboards`;
 export const backend_compile_url = `${backend_baseurl}/v1/compile`;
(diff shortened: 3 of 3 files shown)


yanfali:
This is cool, but it doesn't work out of the box. Try yarn serve there is no default. It really needs a default value.

yanfali:
My preference is to use principle of least astonishment and not change the existing behavior. Make it default to the development if no value is set.

zekth:
> This is cool, but it doesn't work out of the box. Try yarn serve there is no default. It really needs a default value.

Tried and works.
Also :
```
Zekth@DESKTOP-1E2EMMB MINGW64 /c/temp/qmk_configurator (fix-413)
$ echo $VUE_APP_API_URL

```
Nothing prompted. But i can do a fallback in the code.

yanfali:
ok, it was probably my running instance that wasn't picking up change. Restart and it seems to be working now.

yanfali:
> > This is cool, but it doesn't work out of the box. Try yarn serve there is no default. It really needs a default value.
> 
> Tried and works.
> Also :
> 
> ```
> Zekth@DESKTOP-1E2EMMB MINGW64 /c/temp/qmk_configurator (fix-413)
> $ echo $VUE_APP_API_URL
> ```
> 
> Nothing prompted. But i can do a fallback in the code.

I'm also on a mac, so it could have just been a brain fart.

yanfali:
@noroadsleft can you test this on your dev env too? I would like 3rd party confirmation

PR #521, opened 2019-10-05
Update i18n readme with dropdown instructions

kbravh:
I've added instructions for adding a new language to the settings panel dropdown.


yanfali:
Thanks for this. I'm about to change this to make it dramatically easier, so hold off for a day or so.

yanfali:
#522 should make this a lot simpler

kbravh:
Alright I've updated the instructions based on #522! 😁

PR #526, opened 2019-10-06
Fixed bug with spinner where text was appearing gray.

KartikSoneji:
Fixed .spinner css class so that the child classes (.message and .status) would actually have the style applied to them.

## Before
image

## After
image

### DIFF
--- a/src/scss/themes.scss
+++ b/src/scss/themes.scss
@@ -202,8 +202,8 @@ html {
       border-color: #000;
     }
     .spinner {
-      & .message,
-      & .status {
+      .message,
+      .status {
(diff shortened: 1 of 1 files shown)


yanfali:
I think this one crept in because of the dark mode. Can you test whether the & is still needed?

KartikSoneji:
OK, just checked, the `&` is not required.

noroadsleft:
Thanks for this! I noticed it a while back but hadn't had time to track it down, and it was bothering me quite a bit.

KartikSoneji:
You are welcome.

PR #530, opened 2019-10-08
Minor doc change

sanyamdogra:
Used `this` instead of `that` for consistency in error list.


yanfali:
Thanks for your suggestion, but I think we'll keep the wording we have.

Issue #188, opened 2018-08-27
Provide a way to rearrange layers


skullydazed:
The way layers work in QMK is not compatible with the naive model most people approach QMK with. It's really easy to define layers in the wrong order due to that. We should provide a way to rearrange layers so that people don't have to recreate their layers to move them around.

tobster-de:
One could work around this with export/import.

yanfali:
@tobster-de good suggestion, but I have a more UI friendly idea in mind. This is one of my next tasks.

PR #532, opened 2019-10-09
Layer swapping proposal

wojciechgabrys:
I went ahead and added a layer swapping functionality, that allows users to move current layer to a new position (and put target layer in places of moved one).

Let me know if that's something you'd find useful/worth merging in – if yes, I'll add stuff like layer boundry checking and fill in translations.

layer-swapping

### DIFF
--- a/src/store/modules/keymap.js
+++ b/src/store/modules/keymap.js
@@ -275,6 +275,23 @@ const mutations = {
     }
     state.layer = newLayer;
   },
+  swapLayers(state, targetLayer) {
+    if (parseInt(targetLayer) === state.layer) {
+      return;
+    }
+
+    if (state.keymap[targetLayer] === undefined) {
+      mutations.initLayer(state, targetLayer);
+    }
+
+    [state.keymap[targetLayer], state.keymap[state.layer]] = [
+      state.keymap[state.layer],
+      state.keymap[targetLayer]
+    ];
+
+    mutations.changeLayer(state, targetLayer);
+    mutations.setDirty(state);
+  },
   resetConfig(state) {
     state.config = Object.assign({}, state.defaults);
   },
--- a/src/components/LayerControl.vue
+++ b/src/components/LayerControl.vue
@@ -14,7 +14,14 @@
       >{{ layer.name }}</div>
     </div>
     <button
-      class="clear-button"
+      class="layer-action-button"
+      title="Swap Layers"
+      @click="swap"
+    >
+      <font-awesome-icon icon="random" size="lg" fixed-width />
+    </button>
+    <button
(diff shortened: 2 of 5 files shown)


wojciechgabrys:
(Please disregard the failed prettier check – I'll iron it out before the final commit)

yanfali:
that's pretty awesome. However, it's not consistent with how key swapping works. So I'd prefer to make it drag and drop. I just haven't gotten around to it.

src/store/modules/keymap.js
```diff
@@ -275,6 +275,23 @@ const mutations = {
     }
     state.layer = newLayer;
   },
+  swapLayers(state, targetLayer) {
+    if (parseInt(targetLayer) === state.layer) {
```
yanfali: parseInt missing base

src/store/modules/keymap.js
```diff
@@ -275,6 +275,23 @@ const mutations = {
...
+      return;
+    }
+
+    if (state.keymap[targetLayer] === undefined) {
+      mutations.initLayer(state, targetLayer);
+    }
+
+    [state.keymap[targetLayer], state.keymap[state.layer]] = [
```
yanfali: not sure that's a good way to do it. You'd probably want to use Vue.set to make sure the reactivity system is kept in sync

wojciechgabrys:
> that's pretty awesome. However, it's not consistent with how key swapping works. So I'd prefer to make it drag and drop. I just haven't gotten around to it.

Just to clarify, you'd like swapping to work by dragging one layer circle over another one, right? If that's the case do you have any preference regarding drag&drop library I should pull in?

And another question: Layer swap should also automatically update layer indexes on keys (like MO/TG/etc.), right?

noroadsleft:
Will fix #188.

yanfali:
> > that's pretty awesome. However, it's not consistent with how key swapping works. So I'd prefer to make it drag and drop. I just haven't gotten around to it.
> 
> Just to clarify, you'd like swapping to work by dragging one layer circle over another one, right? If that's the case do you have any preference regarding drag&drop library I should pull in?
> 
> And another question: Layer swap should also automatically update layer indexes on keys (like MO/TG/etc.), right?

Yes, layer updating would be necessary. We don't use a layer library, we just use native HTML5 drag and drop. Like I said, this is a lot more work than it seems.

wojciechgabrys:
> Yes, layer updating would be necessary. We don't use a layer library, we just use native HTML5 drag and drop. Like I said, this is a lot more work than it seems.

I'll have a stab at it and see what I can do.

yanfali:
I'll be closing this as it's no longer merge-able.

PR #533, opened 2019-10-09
Fix output bug when printing

shelaf:
Fix bug that the empty author and empty notes are not output correctly when printing.


noroadsleft:
Good work! Thanks!

Issue #126, opened 2018-05-22
Big Switch exposes alignment issues


mechmerlin:
When using the Big Switch with it's Big display: https://config.qmk.fm/#/bigswitch/LAYOUT

You get centered text when using keycodes like KC_A or even Backspace, but when you use something like MO, its centered but moved to the very top. They should all be centered.

wojciechgabrys:
Proposed a fix in https://github.com/qmk/qmk_configurator/pull/534

PR #534, opened 2019-10-09
Fixed key label positioning

wojciechgabrys:
Key labels (in particular ones containing inputs – like ANY, MO, etc.) are now horizontally & vertically centered, regardless of key size (Fixes https://github.com/qmk/qmk_configurator/issues/126)

Fixed key label positioning


noroadsleft:
This is great! Thanks!

PR #542, opened 2019-10-16
Fix */PrintScreen key on converter/xt_usb default keymap

fauxpark:
Relates to qmk/qmk_firmware#7039.

This key is actually the numpad `*` when pressed without Shift or Ctrl:

https://en.wikipedia.org/wiki/File:IBM_Model_F_XT.png


ravijyoti3:
Does it work like printscreen key?

noroadsleft:
Thanks!

PR #548, opened 2019-10-20
Remove the border style on the key when we're inside the print view

### LINKED ISSUE #487: when in print mode, deselect key
symptom: the keyboard layout is shared with the configurator, it's detecting that a specific key has been selected and is highlighting that key on each layer. That's a bug. See issue https://github.com/qmk/qmk_firmware/pull/6683 for an example

Tom-Pietri:
Fixes #487
This remove the selected border on when you're on the print view.

I went with a simple css fix to keep the same logic for how we render the normal keyboard and how we render the print version.

I wasn't sure about where I put the css bit so if it needs to be somewhere else tell me

### DIFF
--- a/src/components/PrintKey.vue
+++ b/src/components/PrintKey.vue
@@ -64,4 +64,8 @@ export default {
   height: 0;
   clear: both;
 }
+
+.print-keymap > .keycode-select {
+  border: none !important;
+}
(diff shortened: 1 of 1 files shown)


noroadsleft:
Looks good, thanks!

PR #562, opened 2019-11-03
Remove macOS-specific Volume Keycodes

noroadsleft:
They don't work in Windows/Linux and thereby confuse users. Also updated titles on the macOS Previous/Next Track keycodes (they behave differently than the standard codes do on macOS).


yanfali:
Looks good, but build is failing

noroadsleft:
CI error is a lint issue in an unrelated file. I'll fix it in another PR.

PR #567, opened 2019-11-05
Searchable keycodes

yanfali:
Proof of Concept for searchable keycodes. @noroadsleft could use some styling input

### DIFF
--- a/src/components/Keycodes.vue
+++ b/src/components/Keycodes.vue
@@ -38,6 +38,8 @@
             type="text"
             :placeholder="$t('message.searchKeycodes')"
             v-model="searchFilter_"
+            autocomplete="off"
+            spellcheck="false"
         /></span>
       </div>
       <div class="tab-area">
--- a/src/components/Keycodes.vue
+++ b/src/components/Keycodes.vue
@@ -167,7 +167,7 @@ export default {
 }
 .tabs {
   display: grid;
-  grid-template: auto / repeat(5, minmax(120px, 200px));
+  grid-template: auto / repeat(6, minmax(120px, 200px));
 }
 .tab {
   text-overflow: ellipsis;
--- a/src/components/Keycodes.vue
+++ b/src/components/Keycodes.vue
@@ -29,7 +29,7 @@
           >{{ $t('message.keycodesTab.' + index + '.label') }}</span
         >
         <span class="end-tab"
-          ><input
+          ><font-awesome-icon class="keycode-search-icon" icon="search"/><input
             @focus="stopListening"
             @blur="startListening"
             type="text"
@@ -170,6 +170,11 @@ export default {
 .end-tab {
   grid-column: -1;
 }
+.end-tab input {
+  padding: 3px 7px;
+  border: 1px solid #cdcdcd;
+  border-radius: 4px;
+}
 .tab-area {
   height: 350px;
   padding: 10px 5px;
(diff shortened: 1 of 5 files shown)


yanfali:
> I'm actually 99% okay with how you've styled this. It would be nice if the search field looked more like the other UI elements (border style, padding, etc.). I had another thought though.
> 
> Could we use key descriptions/titles here and also make them searchable, so if a user searches for `Fn`, it could highlight (well, not dim) the `MO()` key?

I think it already uses titles, would be pretty easy to add one more search field. Try it

noroadsleft:
It's a small change:

```diff
diff --git a/src/components/Keycodes.vue b/src/components/Keycodes.vue
index b8384f3f6..fd627b39f 100644
--- a/src/components/Keycodes.vue
+++ b/src/components/Keycodes.vue
@@ -170,6 +170,11 @@ export default {
 .end-tab {
   grid-column: -1;
 }
+.end-tab input {
+  padding: 3px 7px;
+  border: 1px solid #cdcdcd;
+  border-radius: 4px;
+}
 .tab-area {
   height: 350px;
   padding: 10px 5px;
```

src/components/Keycodes.vue
```diff
@@ -116,6 +125,19 @@ export default {
       this.clearTimeout = window.setTimeout(() => {
         store.commit('app/setMessage', '');
       }, 3000);
+    },
+    filterClass(key) {
+      if (this.searchFilter === '' || isUndefined(key.code)) {
```
zekth: Why not `if (!this.searchFilter` ?

yanfali:
```
s=''
""
!s
true
```

yanfali: I think this is counter intuitive and harder to read. It would only be slightly better if it were !!s

yanfali:
@noroadsleft that styling work for you?

Screen Shot 2019-11-23 at 21 59 31

noroadsleft:
I like the styling but I had another thought:

Is there any way the search could modify the tab labels with the number of keycode results under that section?

Mock-up screenshot because the App tab is too narrow:

image

yanfali:
> I like the styling but I had another thought:
> 
> Is there any way the search could modify the tab labels with the number of keycode results under that section?
> 
> Mock-up screenshot because the App tab is too narrow:
> 
> image

Yeah I thought about it. I just kept it simple for now.

noroadsleft:
We can do it later, that's fine.

yanfali:
Screen Shot 2019-11-25 at 12 32 09

@noroadsleft

src/components/Keycodes.vue
```diff
@@ -145,6 +185,17 @@ export default {
   cursor: pointer;
   margin-bottom: -1px;
 }
+.end-tab {
```
noroadsleft: Makes it look like there's a space between the tab name and the result count.

```suggestion
.tab span {
  margin-left: 4px;
}
.end-tab {
```

src/components/Keycodes.vue
```diff
@@ -25,7 +25,7 @@
           v-for="(key, index) in keycodesByGroup"
           :key="index"
           @click="changeActive(index)"
-          :title="index"
+          :title="$t('message.keycodesTab.' + index + '.label')"
```
noroadsleft: This has driven me nuts for weeks. :heart:

noroadsleft:
Thanks!

PR #568, opened 2019-11-06
Expand ControllerTop Menu Space

noroadsleft:
Makes the CSS grid 3 rows by 2 columns instead of 2 rows by 3 columns, with a wider left-hand column.

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -24,6 +24,19 @@
           ref="select"
         ></v-select>
       </div>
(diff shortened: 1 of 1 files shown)


src/components/ControllerTop.vue
```diff
@@ -422,16 +420,17 @@ export default {
 .topctrl-keyboards {
   grid-row: top;
   grid-column: left;
-  justify-self: start;
+  /*justify-self: start;*/
```
yanfali: Delete

PR #583, opened 2019-11-16
Feature: Show or Hide the Terminal Display

noroadsleft:
Adds a toggle control that collapses (and re-expands!) the terminal display.

[Here's a sample video.](https://cdn.discordapp.com/attachments/456176332825427969/645222311330578444/2019-11-16_03-13-13.mp4)


src/components/StatusPanel.vue
```diff
@@ -22,6 +27,21 @@ export default {
...
       });
+    },
+    /**
+     * toggleTerminal keymap. Attempts to load the keymap data from
+     * a predefined known file path.
+     * @param {boolean} isAutoInit If the method is called by the code
+     * @return {object} promise when it has completed
+     */
+    toggleTerminal() {
+      if (document.getElementById('terminal').getAttribute('class')) {
```
zekth: could you do it the vue way? like: https://vuejs.org/v2/guide/class-and-style.html

src/components/StatusPanel.vue
```diff
@@ -32,3 +52,49 @@ export default {
   }
 };
 </script>
+<style>
+#status {
```
zekth: Please move this in style.scss to be compiled during the build process

noroadsleft: Just the `#status` rule, or the whole style sheet?

zekth: the whole sheet

noroadsleft: 53b8d6bc36c692d5896a62dcdc87a0c6e11ed93d

src/components/StatusPanel.vue
noroadsleft: 2aba62ef78e42e1e87323fa0fadc08f8bde78b1a?

zekth:
A last formating command and it'll be ok :)

noroadsleft:
@zekth Yeah, making some polishing tweaks and then will cap off with a linting.

At some point I should track down why Prettier gives me different results than I get from Travis CI.

PR #584, opened 2019-11-16
Improve logging in the application

zekth:
Moved the console.log to a proper logger to intercept everything the same way and have a factory. Currently it's using pino but, because we have a factory we can move to another dependency if needed. It outputs log like this:

Screenshot 2019-11-16 at 16 19 11

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "axios": "0.19",
     "keypress.js": "^2.1.5",
     "lodash": "4.17.15",
+    "pino": "5.13.6",
     "vue": "2.6",
     "vue-i18n": "^8.10.0",
     "vue-js-toggle-button": "^1.3.2",
--- a/src/App.vue
+++ b/src/App.vue
@@ -80,6 +80,8 @@ import SettingsPanel from '@/components/SettingsPanel';
 import { createNamespacedHelpers, mapActions } from 'vuex';
 const { mapState, mapMutations } = createNamespacedHelpers('app');
 import isFunction from 'lodash/isFunction';
+import { logger as mainLogger } from '@/logger';
+const logger = mainLogger.child({ module: 'APP.vue' });
 export default {
   name: 'app',
   components: {
@@ -104,9 +106,11 @@ export default {
     }
   },
   async beforeMount() {
+    logger.debug('beforeMount');
     await this.appLoad();
   },
   mounted() {
+    logger.debug('mounted');
     this.randomPotatoFact();
     this.interval = setInterval(() => {
       this.randomPotatoFact();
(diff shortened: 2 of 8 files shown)


zekth:
Edit: It's not working on bundled app. Working on this

zekth:
I've done some improvement and i'm happy with it. Now it outputs like this:
Screenshot 2019-11-17 at 08 09 18

yanfali:
let's try and get this merged tomorrow

zekth:
Done :)

zekth:
So seems like resolution of the stack and calling file in FF and Chrome are really different and this makes a bit a mess in the code. Do we still keep this logger feature but we delete the part to point the caller line? IMO i'd not mind as the logger is meant to be meaningfull with the infos in it and pointing the file is just an `extra feature`.

yanfali:
> So seems like resolution of the stack and calling file in FF and Chrome are really different and this makes a bit a mess in the code. Do we still keep this logger feature but we delete the part to point the caller line? IMO i'd not mind as the logger is meant to be meaningfull with the infos in it and pointing the file is just an `extra feature`.

I disagree. We spend more time debugging code than writing it. I'm reluctant to lose a standard feature of the current console implementation.

zekth:
> > So seems like resolution of the stack and calling file in FF and Chrome are really different and this makes a bit a mess in the code. Do we still keep this logger feature but we delete the part to point the caller line? IMO i'd not mind as the logger is meant to be meaningfull with the infos in it and pointing the file is just an `extra feature`.
> 
> I disagree. We spend more time debugging code than writing it. I'm reluctant to lose a standard feature of the current console implementation.

So a log formatter should be implemented.

yanfali:
> > > So seems like resolution of the stack and calling file in FF and Chrome are really different and this makes a bit a mess in the code. Do we still keep this logger feature but we delete the part to point the caller line? IMO i'd not mind as the logger is meant to be meaningfull with the infos in it and pointing the file is just an `extra feature`.
> > 
> > 
> > I disagree. We spend more time debugging code than writing it. I'm reluctant to lose a standard feature of the current console implementation.
> 
> So a log formatter should be implemented.

It should work across FF and Chrome in a similar way. I personally don't do that much development on FF but we always support it, because it's good to have options and it prevents Chrome from becoming the next IE.

zekth:
webpack resolution does not work the same way in FF and Chrome. In Chrome resolution is done if the path is prompted in the console, in the case of FF it's not. That's the blocking part.

yanfali:
> webpack resolution does not work the same way in FF and Chrome. In Chrome resolution is done if the path is prompted in the console, in the case of FF it's not. That's the blocking part.

That sucks. I'm loathe to have a feature that only works on one browser.

noroadsleft:
> That sucks. I'm loathe to have a feature that only works on one browser.

I'm undoubtedly biased, as Firefox is my main browser and has been almost exclusively since 2004, but hard agree. We support Chrome and Firefox officially, and both should be supported equally.

PR #585, opened 2019-11-16
add revision in the app

zekth:
Adding the revision in the app. It's not displayed only for developer purpose.

Also is the notification job in the CI still revelant as we do not use gitter anymore ?

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,6 @@
 <template>
   <div id="app" @click="dismiss">
+    <span style="display:none">{{ revision }}</span>
     <div>
       <header>
         <h1>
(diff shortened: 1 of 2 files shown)


yanfali:
Yeah you can I'll the gitter stuff

src/App.vue
```diff
@@ -28,6 +29,7 @@
     <footer>
       <p>{{ $t('message.maintain') }}</p>
       <p>{{ $t('message.hostedOn') }}</p>
+      <p style="font-size:10px">revision: {{ revision }}</p>
```
yanfali:
```suggestion
      <p style="font-size:10px">version: {{ revision }}</p>
```

PR #588, opened 2019-11-17
Docker setup with nginx

skullydazed:
So that we can share build artifacts between different projects (and because it's becoming more popular to develop locally using docker) I setup a docker container for configurator. We will be able to leverage this for both https://github.com/qmk/qmk_web_stack and https://github.com/qmk/qmk_configurator_appliance.


docker_build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# Setup yarn and build qmk_configurator inside the docker container.
+
+set -e
+#set -x
+
+# Install node
+. /root/.nvm/nvm.sh
+nvm install 11.12
```
yanfali: you should be able to use the .nvmrc in the project for this

docker_build.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Get yarn installed correctly
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+apt-get update
```
zekth: this has to be in the dockerfile

docker_build.sh
skullydazed: Apparently that's not enough, as it complains that 11.12 is not installed when I leave that step out.

skullydazed:
> Not terribly thrilled about including an entire copy of nvm.sh in the project, but I guess it'll have to do until someone has time to figure out a different way.

Yeah, all kinds of bad choices here, nothing good. I tried to curl | bash the install script, but it randomly fails. Same if I include the install script. This seemed like the least bad choice, because at least it works every time.

docker_build.sh
skullydazed: Why is that?

zekth: Because those calls can be cached into a docker layer

skullydazed: My understanding is that it's still cached into a docker layer, and that it's a good idea to collapse your Dockerfile into as few layers as needed. Is my understanding wrong?

zekth: Yes but the whole layer will have the apt-get update and all the install of yarn and the project itself

skullydazed: Yes, and that was my intent in arranging it this way. It bundles all the app dependencies together and ensures that stale dependencies don't break the app.

zekth:
> Not terribly thrilled about including an entire copy of nvm.sh in the project, but I guess it'll have to do until someone has time to figure out a different way.

why not just `curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.1/install.sh | bash` ?

ref: https://github.com/nvm-sh/nvm/blob/master/README.md

docker_build.sh
zekth: Can't this source `.nvmrc` ?

yanfali: I had an discussion with skully on this PR and I would hold off reviewing this PR for a little while, while he reworks stuff

zekth:
Could those scripts be in a `script` folder or a `docker` folder? having it in root can be a mess don't you think?

> > Not terribly thrilled about including an entire copy of nvm.sh in the project, but I guess it'll have to do until someone has time to figure out a different way.
> 
> Yeah, all kinds of bad choices here, nothing good. I tried to curl | bash the install script, but it randomly fails. Same if I include the install script. This seemed like the least bad choice, because at least it works every time.

Weird. Like network errors?

skullydazed:
> Weird. Like network errors?

I'm not sure what was going on. I tried to debug it but the nvm code is pretty spaghetti-ish. Something that's part of generating the NVM URL ends up blanking it out, and I wasn't able to track down why or what. I can give it a try again after some of the reworking I talked to yan about.

yanfali:
> > Weird. Like network errors?
> 
> I'm not sure what was going on. I tried to debug it but the nvm code is pretty spaghetti-ish. Something that's part of generating the NVM URL ends up blanking it out, and I wasn't able to track down why or what. I can give it a try again after some of the reworking I talked to yan about.

Looking forward to seeing the new version. My gut tells me that doing dependency resolution at runtime in docker is not the best way forward. Ideally at build time when we bake the container all dependencies are resolved and we just distribute the build artifacts. Hopefully that's also where Skully is aiming for. Other than that, very excited to see this project happen.

docker_build.sh
skullydazed: After talking to yan a bit yesterday we're revising the direction and goal of this PR. I'm going to rework it so that API requests are always proxied through nginx. This eliminates the need to `yarn build` at startup time, so I'll be refactoring this PR in a way that makes this conversation moot.

skullydazed:
This should ready to go. All API requests go through nginx now, and the only stuff in the build script is strictly related to providing a happy environment to nvm.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+        yarn && \
+    rm -rf /var/lib/apt/lists/*
+RUN curl https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.1/install.sh | bash
+
+# Copy files into place
+COPY . /qmk_configurator/
+COPY conf/nginx.conf.in /etc/nginx/nginx.conf.in
+
+# Build configurator
+RUN /bin/bash -i /qmk_configurator/bin/docker_build.sh
```
yanfali: I think this is almost there. Would love to see this use a multi-stage build to get rid of the tooling. What are your thoughts?

yanfali: https://towardsdatascience.com/how-to-build-slim-docker-images-fast-ecc246d7f4a7 relevant article about multi stage. I think this would be relatively easy to add to docker file

yanfali:
@skullydazed can you give instructions on how to start the container? I tried `docker run -it <image>` and it didn't work. Also your recommended build command would also be good.

yanfali:
> @skullydazed can you give instructions on how to start the container? I tried `docker run -it <image>` and it didn't work. Also your recommended build command would also be good.
> 
> Changes look very good, but I wanted to test it locally before approval

OK, I figured it out from the README. I think showing how to build it would be good to, but other than that I'm ready to approve this. Tested on macOS

yanfali:
@skullydazed are you happy with this?

skullydazed:
I'm happy with the code itself. You wanted me to add build instructions to the readme and I haven't had a chance to sit down to do that yet.

yanfali:
@skullydazed let me know when you're done with docs and we can merge this

skullydazed:
I've just added (and tested) instructions for building the container, and rebased from master. This PR should be ready to go.

skullydazed:
Build failure is due to force push causing a broken git command.

bin/docker_run.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# Setup yarn and build qmk_configurator inside the docker container.
+
+set -e
+#set -x
+
+# Set the API URL
+sed "s;%API_URL%;${QMK_API_URL:-https://api.qmk.fm};g" /etc/nginx/nginx.conf.in > /etc/nginx/nginx.conf
```
zekth: Other than using `sed` on the config file why not using built in nginx way to handle env var?
https://web.archive.org/web/20170712003702/https://docs.apitools.com/blog/2014/07/02/using-environment-variables-in-nginx-conf.html#with-perl

PR #594, opened 2019-11-21
Add GitHub ci

zekth:
Adding github ci instead of travis

Also for the deploy we need a gh token

Seems to be 30s faster than travis


.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
...
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Download Build
+        uses: actions/download-artifact@master
+        if: github.ref == 'refs/heads/master'
+        with:
+          name: build
+          path: dist
+      - name: GitHub Pages Deploy
+        uses: zekth/gh-deploy-action@1.0.0
```
yanfali: why does this depend on a personal repo?

.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          dir: dist
+          gh_token: ${{ secrets.GH_TOKEN }}
```
yanfali: I don't have permissions to create one of these. Where do we get these from?

.github/workflows/build.yml
zekth: There is no official repo for pushing to gh-pages.

.github/workflows/build.yml
zvecr: In this case its provided for you <https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token>

zekth: You can create one for the qmk_bot in the security settings of it

zekth: Or like zvecr said you can also use the generated one. I think it will work also

zvecr: Yeah I'm using it with a different action and it seems fine on push master

zekth: I'll change it. Thanks @zvecr

.github/workflows/build.yml
yanfali: That seems odd

yanfali:
@zekth eli5 is this going to work with our existing configuration you think? I'm a little uncomfortable using a fork from your personal account, I think it would be better if qmk forked your repo and just had a pointer? especially for a deployment task to a public service, it's a little bit risky. What if you lost control of the repo?

zekth:
As long as there is not official github page to actions we can fall back to the deploy.sh script. This needs a bit of tune up but will work also

yanfali:
@zekth go ahead and make it work with deploy.sh and we can merge this if you think it'll work as is

.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
...
+          name: build
+          path: dist
+  deploy:
+    needs: ['build','test']
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Download Build
+        uses: actions/download-artifact@master
+        if: github.ref == 'refs/heads/master'
```
zvecr: the job.if covers this case so can be safely removed
```suggestion
```

zekth: Haha good catch

yanfali: let me know tomorrow when you are around and we can try merging this and see what breaks.

zekth:
@zvecr can you review the changes?

yanfali:
@zvecr if you get a chance

PR #597, opened 2019-11-22
Updates for missing layouts

fauxpark:
Changing default Tragicforce68 keymap to use the proper layout macro, and adding default for Percent Canoe (1:1 with current keymap.c).

For https://github.com/qmk/qmk_firmware/pull/7441


public/keymaps/percent_canoe_default.json
```diff
@@ -0,0 +1,21 @@
+{
```
noroadsleft: Canoe actually has a default but it never got renamed when we added the `percent` folder on qmk_firmware.

Could you dump these contents into `public/keymaps/canoe_default.json` and rename that file to this filename instead?

fauxpark: Done 👍

PR #598, opened 2019-11-22
Vue Analytics deletion

### LINKED ISSUE #592: Analytics dependency deletion
Do we get rid of `vue-analytics` as it's not used anymore due to GDPR ?

zekth:
Fix #592


yanfali:
@zekth can you please up vue.config.js too please

PR #601, opened 2019-11-23
Add translation feature from CSV

zekth:
So here here the move from the json translations to CSV.
Now translations are stored in CSV and generated on the fly. Remaining work is to add a watcher on the csv files.

This is way more flexible for translation contributors.

All the build process is in ./scripts/

Also we are storing everything in `message` parent property but it's not meaningfull. Do i delete this path in the object nesting of translations?

This process is used on several of my projects, here is the documentation i've done for one : https://github.com/reef-pi/reef-pi/blob/master/docs/internationalization_contributing.md (which i can adapt for here)

**Note**: src/i18n/langs.json is just the compiled version of the actual translations.
"i18n:base_gen": "node scripts/json-csv.js" is not meant to be kept in the project as it's used only for the migration process


yanfali:
That's a lot of new stuff. Can you add a README to the top of i18n, you'll probably also need to update the docs that Jack wrote with the new process.

zekth:
> That's a lot of new stuff. Can you add a README to the top of i18n, you'll probably also need to update the docs that Jack wrote with the new process.

OK but can you answer also this ?
> Also we are storing everything in `message` parent property but it's not meaningfull. Do i delete this path in the object nesting of translations?

yanfali:
I don't really understand the question. That's an vue i18n requirement.
Look at the docs.

zekth:
> I don't really understand the question. That's an vue i18n requirement. Look at the docs.

It's not. It's not really obvious in the example of vue-i18n.
In the doc it's an example :
```javascript
const messages = {
  en: {
    message: {
      hello: 'hello world'
    }
  }
}
$t('message.hello')
```
Can be:
```javascript
const messages = {
  en: {
      hello: 'hello world'
  }
}
$t('hello')
```

Having the `message` property is not showing any pro because we use it everywhere.

yanfali:
> > I don't really understand the question. That's an vue i18n requirement. Look at the docs.
> 
> It's not. It's not really obvious in the example of vue-i18n.
> In the doc it's an example :
> 
> ```js
> const messages = {
>   en: {
>     message: {
>       hello: 'hello world'
>     }
>   }
> }
> $t('message.hello')
> ```
> 
> Can be:
> 
> ```js
> const messages = {
>   en: {
>       hello: 'hello world'
>   }
> }
> $t('hello')
> ```
> 
> Having the `message` property is not showing any pro because we use it everywhere.

Oh, the documentation is so contradictory on this. You can remove message.

zekth:
Rebased and cleaned everything. Still needs to do the documentation. Also do i create a task to check and list the missing translations in different languages? I think it could help to have it in the CI.

zekth:
Update: Now a loader is setup to have live reload in the project and not to have to generate them in the first place.

Still missing:
- Documentation
- Add reporting for missing translations for CI

zekth:
Ready for review / comment @yanfali

yanfali:
@zekth 

  - I'm having trouble with the following translations, can you verify they are working?
    - Japanese
    - brazilian
    - polish
- new status bar messages need translation they are missing from current set sorry

zekth:
@yanfali rebased and ready for review.

zekth:
Ping me when you have time to review it, otherwise i'd have to rebase again. I'd prefer to do it only once.

yanfali:
> Ping me when you have time to review it, otherwise i'd have to rebase again. I'd prefer to do it only once.

I can review it today.

yanfali:
@zekth awesome work, thanks for this

PR #605, opened 2019-11-27
add default keymap json for heliar/wm1_hotswap

heliarmk:


public/keymaps/heliar_wm1_hotswap_default.json
```diff
@@ -0,0 +1,22 @@
+{
+  "keyboard": "heliar/wm1_hotswap",
+  "keymap": "default",
```
noroadsleft: You can't actually use `default` as the value for `keymap`, because Configurator's API can see the `keymap.c` files that are in qmk_firmware, even though Configurator itself can't do anything with them. In fact, the value for `keymap` cannot match *any* `keymap.c` that exists in qmk_firmware, for the same reason. The API tries to create a `keyboards/{keyboard}/keymaps/{keymap}/keymap.c` file when the compile job completes, but if one already exists the operation fails.

I append a partial commit hash to the `keymap` value to prevent this.

```suggestion
  "keymap": "default_2b318ba",
```

public/keymaps/heliar_wm1_hotswap_default.json
```diff
@@ -0,0 +1,22 @@
...
+  "commit": "95cc82d1a6c061ce7224d8893049dd0598fd66bd",
```
noroadsleft: I use this for internal tracking, but Configurator's API doesn't use them (yet, I hope).

I get this by going to the default keymap's directory – [here, in your case](https://github.com/qmk/qmk_firmware/blob/master/keyboards/heliar/wm1_hotswap/keymaps/default/keymap.c) – and copying the full commit hash. Every once in a while I check qmk_firmware to see if any of the default keymaps have been updated, by comparing the commit hashes again.

```suggestion
  "commit": "2b318ba01b6e504bfe5f300f8459afc849e104b7",
```

noroadsleft:
Thanks!

PR #611, opened 2019-12-03
Fixed keymap for "Y" and "Z" keys for massdrop/alt "default-ish" keymap

natemcintyre:
In the default-ish layout the "Y" and "Z" keys were located in the wrong places. I have updated the layout to fix this. Unsure exactly how to test these changes as I don't have an environment setup for the qmk/qmk_configurator project, but I would fathom a simple swap would do the trick.


noroadsleft:
Thanks!

PR #625, opened 2019-12-13
update kbd67mkiirgb keymaps

moyi4681:


noroadsleft:
Added via 225c1f9eafa3b9fda83841740eb3607ed9b5d8f9 and b69a49738af7ad9388ecdba15a1a4a686595e10f.

Merging this would have created a couple of issues:

- messed up the file history in ways I don't care for (not your fault; this is a Git thing)
- your v2 keymaps didn't have correct filenames (extra spaces before the file extension)
- none of the keymaps updated the commit hashes, which I like updated even if the keycodes don't change

Also, same as on qmk_firmware, if you commit to a branch created from master, you don't need to delete your fork every time you want to update it. [The guidelines here](https://docs.qmk.fm/#/newbs_git_using_your_master_branch) will work for this repo as well.

PR #626, opened 2019-12-18
Fix default nightmare keymap - align and remove extra keys

zvecr:
Found by `TheToteGoat` on Discord, building the default keymap provided at <https://config.qmk.fm/#/nightmare/LAYOUT_default> produces:

```
Compiling: keyboards/nightmare/nightmare.c                                                          [OK]      
Compiling: keyboards/nightmare/keymaps/default_2a948e7/keymap.c                                    keyboards/nightmare/keymaps/default_2a948e7/keymap.c:5:457: error: macro "LAYOUT_default" passed 52 arguments, but takes just 49
```

Deleting all layers, and recreating from scratch allows the keyboard to build successfully.


noroadsleft:
Ah, I see what happened here. The default keymap in qmk_firmware uses `LAYOUT_split`. The current JSON file has 52 keys for Layers 1 and 2, but 49 for Layer 0, because the base layer is the default keymap's base layer, but applied to `LAYOUT_default` instead.

The better way to fix this would be to modify Layer 0 based on `LAYOUT_split`, and change the layout macro used by the keymap on Line 5.

yanfali:
@noroadsleft what do you want to do?

noroadsleft:
Fixed by [c5c519b](https://github.com/qmk/qmk_configurator/commit/c5c519b89526ac9cf15935a059c8f9817239514e?w=1).

PR #628, opened 2019-12-28
Set dirty keymap if layer is cleared, trigger prompt on page refresh/exit if keymap is dirty

### LINKED ISSUE #145: pop an alert when the user navigates away if keymap is dirty.
just like the other checks, give users a chance to save their work.

nya1:
Closes #145 

Prevent accidental keymap data lost.

Changes in this PR:

 * Set dirty keymap if a layer is cleared
 * Added `beforeunload` listener and show confirmation prompt on page refresh/exit if keymap is dirty

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -124,6 +124,8 @@ export default {
     if (isFunction(this.destroyWatcher)) {
       this.destroyWatcher();
     }
+    // remove event listener
+    window.removeEventListener('beforeunload', this.showConfirmationPrompt);
   },
   computed: {
     ...mapGetters('keymap', ['isDirty']),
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -105,6 +105,10 @@ export default {
   async beforeMount() {
     await this.appLoad();
   },
+  created() {
+    // will trigger function before closing/refreshing tab
+    window.addEventListener('beforeunload', this.showConfirmationPrompt);
```
zekth: You have to remove the eventlistenner in the vuejs event `beforeDestroy`

nya1: right, done

zekth: You've put it in `mounted`. Check the lifecycle of vuejs: https://vuejs.org/v2/guide/instance.html#Instance-Lifecycle-Hooks

nya1: The removeListener is inside `beforeDestroy`, please expand the code above

zekth: My bad.

nya1: no problem

PR #629, opened 2019-12-30
Disable snowflake

yanfali:
time to put festivity stuff away

### DIFF
--- a/src/store/modules/app/state.js
+++ b/src/store/modules/app/state.js
@@ -66,7 +66,7 @@ const state = {
     { value: 'pl-PL', label: 'Polski' },
     { value: 'ms', label: 'Bahasa Malaysia' }
   ],
-  snowflakes: true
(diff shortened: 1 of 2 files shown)


src/store/modules/app/state.js
```diff
@@ -66,7 +66,7 @@ const state = {
     { value: 'pl-PL', label: 'Polski' },
     { value: 'ms', label: 'Bahasa Malaysia' }
   ],
-  snowflakes: true
+  snowflakes: false
```
zekth: why keeping this value in the state if the rest of the app don't have reference to it?

noroadsleft: To re-use it next year, maybe? That's what I would do were this my feature.

yanfali: It's still referenced globally on App.vue. this just makes it hidden and inaccessible

PR #634, opened 2020-01-05
Update Japanese translation

shelaf:

PR #627, opened 2019-12-19
It's beginning to a lot like xmas

yanfali:
A little xmas cheer for the holidays.

  - Updated dependencies
  - fixed bugs
    - vue router has changed to a promise
    - left over message in status bar
  - css based snow flakes
  - svg based santa hat

PR #610, opened 2019-12-02
CSS Polishing

noroadsleft:
A bunch of minor changes mostly targeted toward visual cohesion and readability.

PR #640, opened 2020-01-17
fix translation regressions in #601 and resync

shelaf:
Re-apply translation in #627 ,  #610 , #634 to csv files.


noroadsleft:
Thanks!

PR #650, opened 2020-02-06
Changes for electron app

e11i0t23:
The added code adds an input box for processor under compile/load default buttons when run in the [electron app](https://github.com/e11i0t23/qmk_configurator_electron) (not visible any other time). It auto updates the processor based on loaded keyboard using data already being retrieved from the api

Additionally i have removed the autoflash button and changed compile so when in electron after compiling it goes to automatically flash the keyboard

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -72,6 +72,23 @@
           {{ $t('compile.label') }}
         </button>
       </div>
+      <div class="topctrl-processors" v-if="this.electron">
+        <label
+          class="drop-label"
+          :class="fontAdjustClasses"
+          :title="$t('processorInput')"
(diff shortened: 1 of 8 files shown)


src/components/ControllerTop.vue
```diff
@@ -72,6 +82,23 @@
           {{ $t('compile.label') }}
         </button>
       </div>
+      <div class="topctrl-processors" v-if="this.electron">
```
yanfali: this should probably be it's own component

src/components/ControllerTop.vue
```diff
@@ -64,6 +64,16 @@
           {{ $t('loadDefault.label') }}
         </button>
         <button
+          v-if="this.electron"
```
yanfali: should probably be a component

PR #653, opened 2020-02-08
add default keymap for Prime_L V2

holtenc:


holtenc:
Perhaps this layout for configurator should be separate from the original Prime_L because the bottom row key sizes have changed from V1 to V2

public/keymaps/primekb_prime_l_v2_default.json
```diff
@@ -0,0 +1,22 @@
+{
+  "keyboard": "primekb/prime_l_v2",
```
noroadsleft:
```suggestion
  "keyboard": "primekb/prime_l/v2",
```

public/keymaps/primekb_prime_l_v2_default.json
```diff
@@ -0,0 +1,22 @@
...
+  "keymap": "default_8fe29f2",
```
noroadsleft:
```suggestion
  "keymap": "default_42d6270",
```

public/keymaps/primekb_prime_l_v2_default.json
```diff
@@ -0,0 +1,22 @@
...
+  "commit": "8fe29f2784b1c14a75694a908b8a5116d9517b97",
```
noroadsleft:
```suggestion
  "commit": "42d6270f28831e95d1cb9c14a7423d5b1d864d67",
```

holtenc:
that should be done. at least in a round about way.

PR #654, opened 2020-02-11
Load keymaps from subdirectories

noroadsleft:
This is a stripped-down version of #591; it doesn't do any of the layout macro business. Seven hundred keymap files in one folder is getting to be a real PITA to maintain.


yanfali:
I think this is just too much. I would go with something simpler. Just take the first character of every existing path and make that a folder. [0-9a-z] so 38 folders and just drop them in alphabetically into those folders. This is fast and simple string operation to figure out where to search.

PR #662, opened 2020-02-22
Hide Unused Layers

noroadsleft:
Hides the selectors for the layers that are unused in the current keymap.

image

### DIFF
--- a/src/scss/style.scss
+++ b/src/scss/style.scss
@@ -411,16 +411,14 @@ button {
 }
 
 .layers {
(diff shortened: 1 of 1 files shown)


yanfali:
How do you as a layer?

noroadsleft:
And this is why we code review. :man_facepalming: A button for adding a layer might be needed. Otherwise the only way is to initialize one through the layer change keycodes.

yanfali:
what's the status of this?

PR #674, opened 2020-03-07
Candybar: update to address different layouts.

TerryMathews:
Update to reflect project split into lefty and righty subprojects happening in QMK. This should not be merged until QMK PR is also merged, as it will break existing Candybar support in config.qmk.fm.

### DIFF
--- a/public/keymaps/c/candybar_lefty_default.json
+++ b/public/keymaps/c/candybar_lefty_default.json
@@ -1,0 +1,20 @@
+{
+  "keyboard": "candybar/lefty",
+  "keymap": "",
+  "commit": "",
+  "layout": "LAYOUT",
+  "layers": [
+    [
+      "KC_P7",   "KC_P8",   "KC_P9",   "KC_PAST", "KC_ESC",  "KC_Q",    "KC_W",    "KC_E",    "KC_R",    "KC_T",    "KC_Y",    "KC_U",    "KC_I",    "KC_O",    "KC_P",    "KC_DEL",  "KC_BSPC",
+      "KC_P4",   "KC_P5",   "KC_P6",   "KC_PMNS", "KC_TAB",  "KC_A",    "KC_S",    "KC_D",    "KC_F",    "KC_G",    "KC_H",    "KC_J",    "KC_K",    "KC_L",    "KC_SCLN",            "KC_ENT",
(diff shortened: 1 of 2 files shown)


noroadsleft:
Requires qmk/qmk_firmware#8335.

TerryMathews:
@noroadsleft this is ready to merge.

public/keymaps/c/candybar_lefty_default.json
```diff
@@ -0,0 +1,20 @@
+{
+  "keyboard": "candybar/lefty",
+  "keymap": "",
+  "commit": "",
```
noroadsleft:
```suggestion
  "keymap": "default_18bc525",
  "commit": "18bc5254932a842276627d8f055b6fb1d3345230",
```

public/keymaps/c/candybar_righty_default.json
```diff
@@ -1,7 +1,7 @@
 {
-  "keyboard": "candybar",
-  "keymap": "default_0072fdd",
-  "commit": "0072fdd799ffe61bf64f12c23335da3adeb083e5",
+  "keyboard": "candybar/righty",
+  "keymap": "",
+  "commit": "",
```
noroadsleft:
```suggestion
  "keymap": "default_18bc525",
  "commit": "18bc5254932a842276627d8f055b6fb1d3345230",
```

TerryMathews:
Done.

noroadsleft:
Thanks!

TerryMathews:
@noroadsleft how long should it take for the defaults to be available on config.qmk.fm?

noroadsleft:
Normally it should only be something like five minutes or so, but the backend is running really slowly today, and it seems the update hasn't deployed.

Sorry for the long response, I've been in and out all day.

TerryMathews:
@noroadsleft it's still not loading them as of 8:23am eastern. If I use the URL import pointed at the raw file in the repo, it loads fine.

Has the server still not refreshed, or is there something wrong with the naming I used?

noroadsleft:
Nah, your files are fine. Near as I can tell is what's happening is:

- the server that provides the API data for Configurator is having high response times (doesn't respond quickly)
- the high response times mean the Configurator's UI is slow to load
- the slow page load times means some HTML elements don't exist (aren't loaded into the DOM) when the tests kick off
- the tests not finding the HTML elements means some of the tests fail
- the failing tests return error codes that mean something didn't work
- the error codes noting something didn't work means the CI doesn't deploy the updated version

I've been restarting the test/deploy job periodically but it hasn't passed yet. Also some of the CI infrastructure has changed as far as the containers, and it may have broken our test tooling.

noroadsleft:
We extended the timeout duration for the tests earlier and this is working now.

PR #678, opened 2020-03-10
'\n' doesn't work with CSV, use quoted strings

shelaf:
'\n' in the message must be parsed as a new line.
https://github.com/qmk/qmk_configurator/blob/master/src/i18n/en.csv#L15


yanfali:
@zekth

zekth:
Regarding csv specs it's a problem of csv file not parsing process:
```
statsTemplate,"
Loaded {layers} layers and {count} keycodes. Defined {any} Any key keycodes
"
```

PR #694, opened 2020-03-29
Refactoring Tasks

yanfali:
@zekth some overdue clean up

- catch some errors
  - routing to same keyboard while loading JSON
  - keycount getter is invalid while viewing preview
- remove store references in ControllerBottom
- initialize icons in a more consistent way
- download mime type was incorrect for JSON

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -238,8 +238,7 @@ export default {
             this.updateLayout(data.layout);
             let promise = new Promise(resolve =>
               store.commit('keymap/setLoadingKeymapPromise', resolve)
-            );
-            promise.then(() => {
+            ).then(() => {
               this.updateKeymapName(data.keymap);
               const stats = load_converted_keymap(data.layers);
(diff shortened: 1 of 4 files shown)


src/components/ControllerTop.vue
```diff
@@ -238,8 +238,7 @@ export default {
             this.updateLayout(data.layout);
             let promise = new Promise(resolve =>
               store.commit('keymap/setLoadingKeymapPromise', resolve)
```
zekth: Waiting for a commit wouldn't be in an action? it's more the vue design?

yanfali:
It's used as a waiting mechanism in a different part of the application.
This sets the value only.

yanfali:
Good catch. I'll look into it.

yanfali:
> Key counter doesn't work at all.
> 
> image

fixed

src/components/ControllerTop.vue
yanfali: this isn't an action, it's a signaling mechanism.

PR #698, opened 2020-04-03
Add default keymap for bear_face

chemicalwill:
* Adds configurator support for bear_face PCB

* Changed base layer change macros to TO(layer)


noroadsleft:
Thanks!

PR #703, opened 2020-04-13
Issue 692 - Update UI changes for API improvements

yanfali:
UI Changes
  - removing button
  - adding help button
  - improved tooltips
- templating the JSON export format using the API

OK I think this is ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "core-js": "^3.4.4",
     "keypress.js": "^2.1.5",
     "lodash": "4.17.15",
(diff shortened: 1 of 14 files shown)


yanfali:
let me verify the whole download.json problem

yanfali:
OK ready to go. I investigated the keymap naming issue you brought up. It takes the default name from the default json, so if you want the keyboard and layout in the name just fix up the json files. It does the right thing if you delete the name specified.

PR #710, opened 2020-04-19
Add dynamic width to ANY input

christophernewton:
Hi there,

**Whats in this PR**
An update to make the input on the any key dynamic to accomodate large combinations of strings e.g.
Kapture 2020-04-19 at 12 07 57

**Why**
Previously the input would remain static and this made it hard to look at all the content in the input  a small workaround to this was to use the json import so that I could see the whole string.

Ive only applied this update to the any key and have left the other inputs which use a similar layout such as layer selection

Tests are all passing, if theres any amendments / updates Id be happy to chat through them!

(Also with self isolation I have a bit more time so if there are any other quality of life improvements id be happy to help contribute where I can)

Thanks

### DIFF
--- a/public/keymaps/e/ergodox_ez_default.json
+++ b/public/keymaps/e/ergodox_ez_default.json
@@ -28,7 +28,7 @@
       "KC_TRNS", "KC_EXLM", "KC_AT",   "KC_LCBR", "KC_RCBR", "KC_PIPE", "KC_TRNS",
       "KC_TRNS", "KC_HASH", "KC_DLR",  "KC_LPRN", "KC_RPRN", "KC_GRV",
       "KC_TRNS", "KC_PERC", "KC_CIRC", "KC_LBRC", "KC_RBRC", "KC_TILD", "KC_TRNS",
-      "EPP_RST", "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_TRNS",
+      "EEP_RST", "KC_TRNS", "KC_TRNS", "KC_TRNS", "KC_TRNS",
                                                              "RGB_MOD", "KC_TRNS",
                                                                         "KC_TRNS",
                                                   "RGB_VAD", "RGB_VAI", "KC_TRNS",


yanfali:
that is slick. thanks.

src/components/LayerKey.vue
```diff
@@ -83,9 +84,11 @@ export default {
     blur() {
       this.startListening();
       this.setSelected(undefined);
+      this.hasFocus = false;
```
yanfali, line 87: why does this key need focus and none of the others?

src/components/LayerKey.vue
```diff
@@ -83,9 +84,11 @@ export default {
...
     },
     focus() {
       this.stopListening();
+      this.hasFocus = true;
```
yanfali, line 91: ditto

src/components/LayerKey.vue
christophernewton: The core part of how this works is that the input needed to overflow outside the bounds of the key, to do this I used the [pseudo selector :focus](https://github.com/qmk/qmk_configurator/pull/710/files/2b76eda2f5dd6c9eb646fb5b37b32946f0d054d5#diff-b86aa89eb58031c905f46ced082fdd9fR604) to set the input to absolute. Because of this I needed to make sure the titles didnt move from their position so by adding this I could add padding to the bottom of the key title to give the illusion that it didnt move. Without this the title "Any" disappears because each key is display:flex;

I do agree this is a little out of the box but I also wanted to approach this without refactoring any other components.

src/components/LayerKey.vue
christophernewton: The same position absolute trick is applied here (because the scss is global and not scoped) so to avoid refactoring and renaming classes I decided to give it the same treatment but just not enable the input to expand as traditionally the layer input would be 0,1 or 2 chars wide which works already.

src/components/LayerKey.vue
yanfali: OK, thanks for eli5. this makes a lot more sense now.

yanfali:
@noroadsleft any comments? otherwise I'll merge.

PR #716, opened 2020-04-21
Add dark mode if user OS has opted in.

christophernewton:
Hi again,

**What is in this PR**

On init of the app I can see vuex will default `false` for dark mode, this little update will add dark mode if your OS has it turned on with the caveat being that local storage must not already exist and set to `darkmode: false`

Here is a quick example of me clearing the configurator localstorage and doing a hard reload
Kapture 2020-04-21 at 16 31 46

On load using the `window.matchMedia()` function it will enable or disable dark mode for the user automatically if they have opted in for their OS

**Why**

Im a developer and love darkmode ;)

**Tests**

I updated the localStorage spec because it had hard coded expectations that the flag is set to false by default after clearing localstorage. I would be happy to make this more dynamic if you would like but for now I forced the color scheme in cypress to return false.

Chris 👍


yanfali:
@zekth this was your feature. care to weigh in?

yanfali:
Looks minimally intrusive. I like it overall.

src/store/modules/app/state.js
```diff
@@ -5,9 +5,13 @@ import {
 } from '@/store/localStorage';
 
 function setDefaultConfiguratorSettings() {
+  // detect if OS supports dark mode and set as default
+  let osDarkMode =
```
zekth, line 9:
```suggestion
  const osDarkMode =
```

christophernewton:
> Other than this keyword it's a great addition. I wasn't aware of this API.

Its a handy little api to tap into for feature detection, the property its actually targeting is a [css media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) that was introduced a fairly recently (which is why it will fall back to false if it doesnt pick it up)

PR #717, opened 2020-04-22
Translate Keycodes on Keymap Import

noroadsleft:
This feature polishes the Configurator UI's understanding of what keycode is being used when the imported keymap uses an alias keycode that Configurator doesn't understand natively.

## Example

`_______` is a QMK alias for the Transparent keycode `KC_TRANSPARENT` keycode, which Configurator's UI code understands as `KC_TRNS`. When a keymap is imported, Configurator will remap the alias to its natively-understood keycode:

image

## To-Do

- [x] Revert the "bugged" Gherkin keymap

## Things for a later PR?

- Rewrite the default keymaps to use the native keycodes
- Switch the natively-understood keycodes to be the most common variants present in qmk_firmware
- ???


src/longFormKeycodes.js
```diff
@@ -0,0 +1,235 @@
+const longFormKeycodes = {
```
yanfali, line 1: use the `const export`

src/longFormKeycodes.js
```diff
@@ -0,0 +1,235 @@
...
+  UNICODE_MODE_FORWARD: 'UC_MOD',
+  UNICODE_MODE_REVERSE: 'UC_RMOD',
+  UNICODE_MODE_OSX: 'UC_M_OS',
+  UNICODE_MODE_LNX: 'UC_M_LN',
+  UNICODE_MODE_WIN: 'UC_M_WI',
+  UNICODE_MODE_BSD: 'UC_M_BS',
+  UNICODE_MODE_WINC: 'UC_M_WC'
+};
+
+export { longFormKeycodes };
```
yanfali, line 235: then you can remove this line

src/longFormKeycodes.js
noroadsleft: It was `export const`, but thanks. :+1:

PR #718, opened 2020-04-24
breaking Rebound into two revs

Rossman360:


noroadsleft:
I actually have a branch locally that updates this in conjunction with some other related updates:

- Adds `montsinger/rebound` to the Exclusion List (prevents it from being selected as a build target in the UI; as it's the same as `montsinger/rebound/rev1`).
- Anyone who imports a keymap that sets `keyboard` to `montsinger/rebound` will be redirected to `montsinger/rebound/rev1`.
- Updates the commit hashes and the keymap name. The Configurator's API doesn't have the keymap name collision issue anymore, so there isn't any need to set `keymap` to something other than `default`. Configurator also now ignores the keymap name when the internal default keymap is loaded (Keymap Name is `null` in the UI).
- I also just added a feature where Configurator is now less particular about which keycodes are used in keymaps that are imported, which means its UI code no longer cares whether you use `KC_TRANSPARENT`, `KC_TRNS`, or `_______` – Configurator will render all those codes identically. `montsinger/rebound/rev2` gets to be the lucky dog, as the first keymap to use this feature. :)

My PR will supersede this one, so I'll be closing this when I get that PR posted, but I'll tag you on the PR.

Thanks for keeping up with updates, though. I appreciate it.

PR #721, opened 2020-04-25
Update keymaps for montsinger/rebound

noroadsleft:
## update keymaps for montsinger/rebound

- relocate: montsinger/rebound -> montsinger/rebound/rev1
- add keymap for rev2
- update exclusion list and redirect table

Supersedes #718.


noroadsleft:
@Rossman360 So the components of this that are different from your PR are:


## Metadata updates

Specify `rev1` in the rev1 keymap, update the commit hash for tracking, and remove the commit hash suffix from the keymap name.

https://github.com/qmk/qmk_configurator/blob/a60968f5d2d64e9a7de46eccb68cee37f8b94f8d/public/keymaps/m/montsinger_rebound_rev1_default.json#L2-L4


## Tri-Layer Emulation

I realized when I was going through keymap updates recently that you use the tri-layer feature in your default keymap (Lower + Raise = Adjust), which I emulate by using `MO(5)` on your Lower and Raise layers here:

https://github.com/qmk/qmk_configurator/blob/a60968f5d2d64e9a7de46eccb68cee37f8b94f8d/public/keymaps/m/montsinger_rebound_rev1_default.json#L29 and https://github.com/qmk/qmk_configurator/blob/a60968f5d2d64e9a7de46eccb68cee37f8b94f8d/public/keymaps/m/montsinger_rebound_rev1_default.json#L35


## White-Space and Alias keycodes on the rev2 keymap

`public/keymaps/m/montsinger_rebound_rev2_default.json` only differs from your version in white space, and use of alias keycodes – `S(KC_NUHS)` instead of `LSFT(KC_NUHS)`, `_______` instead of `KC_TRNS`.


## Exclude `montsinger/rebound` as a build target.

We discovered in past that not requiring the user to know which revision of a PCB they had led to issues where users would build a keymap that Configurator would interpret as, for instance, `make planck/rev5:user_keymap` and then attempt to flash it to a Planck rev6 PCB, and get errors. This change requires the user to know which Rebound revision they have (though this should be more obvious given rev2 has one more key).

https://github.com/qmk/qmk_configurator/pull/721/files#diff-fcf8502f64bc16b0bf8e36376e1b786aR512


## Redirect the build target for montsinger/rebound (rev1) imported keymaps

This makes it so Configurator treats keymap JSONs that use `"keyboard": "montsigner/rebound"` as having `"keyboard": "montsigner/rebound/rev1"` set instead.

https://github.com/qmk/qmk_configurator/pull/721/files#diff-3ee91f2281d5c6be3344e183462ff900R80-R82

Rossman360:
eeeeee, thank you so much for being awesome

noroadsleft:
No, thank *you!*

PR #722, opened 2020-04-26
Simplify BaseKeymap setSize methods

kbrock:
While going through the keymap code to add in rotation, this code struck me as unneeded.

- `setSize` is defined in the base class
- `calcMax` is not used
- the double negative subtraction is a bit much for me.

This aims to not change functionality.
If you want me to remove any one of the commits, just holler.

Thanks for all the help and the great tool

### DIFF
--- a/src/components/VisualKeymap.vue
+++ b/src/components/VisualKeymap.vue
@@ -155,27 +155,7 @@ export default {
 
       // eslint-disable-next-line no-console
       this.profile && console.time('layout::scale');
-      const layout = this.layouts[newLayout];
-      const max = layout.reduce(
-        (acc, pos) => {
-          let _pos = Object.assign({ w: 1, h: 1 }, pos);
-          const coor = this.calcKeyKeymapPos(_pos.x, _pos.y);
-          const dims = this.calcKeyKeymapDims(_pos.w, _pos.h);
-          acc.x = Math.max(acc.x, coor.x + dims.w);
-          acc.y = Math.max(acc.y, coor.y + dims.h);
-          return acc;
-        },
-        {
-          x: 0,
-          y: 0
-        }
-      );
-      if (max.x > this.defaults.MAX_X) {
-        this.resizeConfig(max);
-        max.x *= this.config.SCALE;
-        max.y *= this.config.SCALE;
-      }
-      this.setSize(max);
+      this.setSize(this.calculateMax(newLayout));
       // eslint-disable-next-line no-console
       this.profile && console.timeEnd('layout::scale');
     }
--- a/src/components/VisualKeymap.vue
+++ b/src/components/VisualKeymap.vue
@@ -135,10 +135,6 @@ export default {
           return BaseKey;
       }
     },
-    setSize(max) {
-      this.width = max.x;
-      this.height = max.y;
-    },
     recalcEverything(newLayout) {
       // eslint-disable-next-line no-console
       this.profile && console.time('layout::reset');


src/components/BaseKeymap.vue
```diff
@@ -4,12 +4,8 @@ export default {
   methods: {
     calcKeyKeymapDims(w, h) {
       return {
-        w:
-          w * this.config.KEY_X_SPACING -
-          (this.config.KEY_X_SPACING - this.config.KEY_WIDTH),
-        h:
-          h * this.config.KEY_Y_SPACING -
-          (this.config.KEY_Y_SPACING - this.config.KEY_HEIGHT),
+        w: (w-1) * this.config.KEY_X_SPACING + this.config.KEY_WIDTH,
```
yanfali, line 7: try using a really big keyboard layout. this code doesn't do what you think it does. by using a constant instead of the configuration block you have probably introduced a bug. It's a scaler.

yanfali: So, I'm trying to understand what the advantage of this is? Maybe I'm too dense, but when I punch numbers into a spread sheet they no longer seem equivalent.

w = 3, KEY_X_SPACING = 40, KEY_WIDTH = 45.
Old formula = (3 * 40) - (40-45) = 157
New formula = 2 * 40 + 45 = 125

Unless I'm missing something, this holds true for all values of w.

yanfali:
Couple of other issues. Linting isn't passing.

And I'm seeing this error in the console which doesn't exist in production

Screen Shot 2020-04-26 at 17 26 00

Not sure what's going there.

yanfali:
> Couple of other issues. Linting isn't passing.
> 
> And I'm seeing this error in the console which doesn't exist in production
> 
> Screen Shot 2020-04-26 at 17 26 00
> 
> Not sure what's going there.

easy way to reproduce. Type 'hasu' into the configurator keyboard drop down and hit enter.

src/components/BaseKeymap.vue
kbrock: I read:

```python
rslt = w * this.config.KEY_X_SPACING - (this.config.KEY_X_SPACING - this.config.KEY_WIDTH)
rslt = w * this.config.KEY_X_SPACING - this.config.KEY_X_SPACING + this.config.KEY_WIDTH
rslt = (w - 1) * this.config.KEY_X_SPACING + this.config.KEY_WIDTH
```
That seems the same to me

```python
w = 3, KEY_X_SPACING = 45, KEY_WIDTH = 40.
# before
rslt = w * this.config.KEY_X_SPACING - (this.config.KEY_X_SPACING - this.config.KEY_WIDTH)
Old formula = (3 * 45) - (45-40) = 130

# after
rslt = (w - 1) * this.config.KEY_X_SPACING + this.config.KEY_WIDTH
New formula = 2 * 45 + 40 = 130
```

yanfali: I see. I punched in the wrong value. My bad.

kbrock:
I'll fix the lint issues. in the mean time, I'm not married to the width/height change. So if the old code is easier for you to read, then lets remove this commit from the PR.

I'm working towards a prototype that supports key rotation. That is something I care about. This was just some code I was trying to understand as I worked through the code base.

yanfali:
@kbrock @skullydazed  and looked at the rotation implementation on KLE and actually I think I could probably retrofit something as soon as the data was put back info info.json. I don't think it's that difficult technically. It's just a question of when.

For your PR to make further progress, you'd need to find the cause for the errors.

src/components/VisualKeymap.vue
```diff
@@ -166,27 +162,7 @@ export default {
...
-          y: 0
-        }
-      );
-      if (max.x > this.defaults.MAX_X) {
-        this.resizeConfig(max);
-        max.x *= this.config.SCALE;
-        max.y *= this.config.SCALE;
-      }
-      this.setSize(max);
+      this.setSize(this.calculateMax(this.layout));
```
kbrock, line 165:
```suggestion
      this.setSize(this.calculateMax(newLayout));
```

kbrock:
totally did not see the lint errors in there. `yarn run prettier:check --write` is pretty sweet

lets hope the tests now pass

yanfali:
Not sure if we need this anymore.

kbrock:
yea, the `w`/`h` stuff was not needed from your uh, uw changes.

Deduping the `setSize` is not necessary if you're planning on dropping the setSize. Didn't think the `setSize` was going away any time soon.

I can close this if you want.

yanfali:
Thanks, I think we should close this

PR #728, opened 2020-05-03
Move per key height and width calculations back into CSS

yanfali:
Take advantage of CSS Custom Variables and CSS calc to move height and width calculations into CSS where they really belong rather than per DOM key element.

 - ISO CSS contributed by @kbrock thanks so much for this work!

### Pros:
 - remove explicit height and width in each key style
 - reduces parsing load on the DOM
 - allows overriding through use of classes. Styles have highest specificity

### Cons:
 - pushes some work in the CSS parser
 - may be somewhat difficult to understand

### DIFF
--- a/src/components/BaseKey.vue
+++ b/src/components/BaseKey.vue
@@ -34,6 +34,65 @@ let substitute = Object.assign(
   colorways.platformIcons(window.navigator.platform)
 );
 
+const _getUnitClass = (unit, rotated) => {
+  if (rotated) {
+    switch (unit) {
+      case 2:
+        return 'k2uh';
+      case 1.25:
+        'k125uh';
+        return 'k125uh';
+      case 1.5:
+        return 'k15uh';
+      case 1.75:
+        return 'k175uh';
+      default:
+        return 'kiso';
+    }
+  }
+  switch (unit) {
+    case 1:
+      return 'k1u';
+    case 1.25:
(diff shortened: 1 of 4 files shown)


yanfali:
> 
> Is there any way we could get this to support arbitrary sizes?

We might be able to generate it from scss, it would require some parsing magic. @zekth any ideas? Tbh, all those keyboards are 1 offs and fixing them likely fixes things for a while. As for the rounding/typos errors, those should just be fixed.

noroadsleft:
> As for the rounding/typos errors, those should just be fixed.

I failed to specify, but yes, that is my intention.

yanfali:
@noroadsleft ready for review

yanfali:
Screen Shot 2020-05-17 at 18 32 32
Screen Shot 2020-05-17 at 18 32 50
Screen Shot 2020-05-17 at 18 33 15
Screen Shot 2020-05-17 at 18 33 42
Screen Shot 2020-05-17 at 18 34 05

noroadsleft:
Feature looks good, but is 909492cf89acf88260619f916fe9b5fc14fa6a41 meant to be on this branch?

yanfali:
@noroadsleft good catch. fixed

kbrock:
looks very good

noroadsleft:
Thanks!

PR #731, opened 2020-05-15
Murcielago default layout

elagil:
After merging the board to qmk, I would like to add this default layout. I chose a commit ID that was yet unused. I hope that it is ok like this. It looks a bit messy, due to the long keycodes.

How can I make sure that only murcielago/rev1 appears in the configurator drop-down menu, and not murcielago alone (without rev1)? The latter does not build, which makes sense.

### DIFF
--- a/public/keymaps/m/murcielago_rev1_default.json
+++ b/public/keymaps/m/murcielago_rev1_default.json
@@ -1,0 +1,36 @@
+{
+  "keyboard": "murcielago/rev1",
+  "keymap": "default_6a1a08a1",
+  "commit": "6a1a08a162ab1d03e823cb21df3f2e427bd1508c",
+  "layout": "LAYOUT",
+  "layers": [
+    [
+      "KC_EQL",   "KC_1",           "KC_2",    "KC_3",   "KC_4",         "KC_5",                       "KC_6",          "KC_7",   "KC_8",   "KC_9",         "KC_0",   "KC_MINS",
+      "KC_LBRC",  "KC_Q",           "KC_W",    "KC_E",   "KC_R",         "KC_T",                       "KC_Y",          "KC_U",   "KC_I",   "KC_O",         "KC_P",   "KC_RBRC",
+      "KC_DEL",   "KC_A",           "KC_S",    "KC_D",   "KC_F",         "KC_G",                       "KC_H",          "KC_J",   "KC_K",   "KC_L",         "KC_SCLN","KC_QUOT",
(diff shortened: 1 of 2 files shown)


public/keymaps/m/murcielago_rev1_default.json
```diff
@@ -0,0 +1,36 @@
+{
+  "keyboard": "murcielago/rev1",
+  "keymap": "default_6a1a08a1",
+  "commit": "6a1a08a162ab1d03e823cb21df3f2e427bd1508c",
```
noroadsleft, line 4:
```suggestion
  "keymap": "default",
  "commit": "fadd3cb4617fe7e48c802c4470a50df36e6c5109",
```

- The `keymap` key no longer has the name collision issue in the API. You can use `default` as the value here.
- I get the commit hashes by running `git log -1 --pretty=format:"%H" -- keyboards/<keyboard>/keymaps/default/keymap.c` from QMK's current `master` branch, e.g.
  ```shell
  git checkout master
  git pull upstream master
  git log -1 --pretty=format:"%H" -- keyboards/murcielago/rev1/keymaps/default/keymap.c
  ```

noroadsleft:
> How can I make sure that only murcielago/rev1 appears in the configurator drop-down menu, and not murcielago alone (without rev1)? The latter does not build, which makes sense.

That's caused by using the `DEFAULT_FOLDER` rule on the qmk_firmware side.

You'll need to go to `/src/jquery.js` and insert `'murcielago',` at Line 513.

public/keymaps/m/murcielago_rev1_default.json
elagil: Ok thanks, I though that it was just chosen randomly!

elagil:
> You'll need to go to `/src/jquery.js` and insert `'murcielago',` at Line 513.

Cool, I will do that.

elagil:
That should be it!

noroadsleft:
Thanks!

elagil:
Thank you! I noticed that, despite the exclusion, murcielago still appears in the list. The keymap for rev1 works fine, though.

noroadsleft:
> Thank you! I noticed that, despite the exclusion, murcielago still appears in the list. The keymap for rev1 works fine, though.
> 
> EDIT: Sorry, it vanished now! All good.

It takes about 5-10 minutes after a merge before the changes are reflected in Configurator, and Configurator also only ever references the `master` branch of this repo. The `jquery.js` edit is what excludes the "dead" build target from the list, but it's gotta be deployed. :)

PR #732, opened 2020-05-15
Murcielago: change keymap according to qmk defaults

elagil:
Just a small change to the default keymap.
The previous RESET keycodes don't make a lot of sense if only one half is connected, as they cannot be reached in that case.

Also, make SHIFT transparent on function layers, or text cannot be selected in combination with arrows keys.

### DIFF
--- a/public/keymaps/m/murcielago_rev1_default.json
+++ b/public/keymaps/m/murcielago_rev1_default.json
@@ -5,32 +5,32 @@
   "layout": "LAYOUT",
   "layers": [
     [
-      "KC_EQL",   "KC_1",           "KC_2",    "KC_3",   "KC_4",         "KC_5",                       "KC_6",          "KC_7",   "KC_8",   "KC_9",         "KC_0",   "KC_MINS",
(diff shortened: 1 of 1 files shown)


elagil:
I will add further improvements and make a unified pull request later

yanfali:
Should we close this?

elagil:
I will update this soon, so I wouldn't close it.

noroadsleft:
@elagil Are you planning on updating the default keymap that's on qmk_firmware? Near as I can see at a glance this keymap already matches the keymap over there and doesn't need to be updated?

elagil:
It will be the same keymap after my PR on the qmk_firmware side passes: https://github.com/qmk/qmk_firmware/pull/9363

elagil:
> Change the value here to `e2699773879f17e70e05dd043f2231786c6a61ad`.

Sorry, I forgot about that!

noroadsleft:
Thanks!

elagil:
Thanks again for merging!

PR #748, opened 2020-06-16
Issue 746: toggle default clear keycode

yanfali:
A little more involved that I expected but this allows toggling of
the default behavior for clearing a layer. There's one downside,
you have to toggle it back to actually remove a layer if you are
in KC_TRNS mode, but this is an expert mode thing...


yanfali:
@zekth weigh in please since I'm messing with local store

yanfali:
By design you may not clear the base layer (0) as anything by KC_NO as there's no where else to go.

PR #753, opened 2020-06-18
Add some diagnostic messages for unknown layouts

yanfali:
This prevents an all out crash which isn't helpful, and gives you a diagnostic about bad defaults that reference layouts that no longer exist. Ideally this should have been remapped.


yanfali:
> Looks good and works, but I wonder if this would be more useful if it output to the terminal display in addition to or instead of the browser console.

That's a good idea.

noroadsleft:
Thanks!

PR #759, opened 2020-06-22
Fix custom key sizes

fauxpark:
The keys on the `bigseries` keyboards were being given the class `k4u`, but they are actually 4x4u. Additionally, the sizing calculations for custom sized keys needed to take the key spacing into account.

image

### DIFF
--- a/src/components/BaseKey.vue
+++ b/src/components/BaseKey.vue
@@ -35,7 +35,10 @@ let substitute = Object.assign(
 );
 
 const _getUnitClass = (unith, unitw) => {
-  if (unith >= unitw && unith !== 1) {
+  if (unith == unitw && unith > 1) {
+    return 'custom';
+  }
+  if (unith > unitw || unith < 1) {
     if (unith === 2 && unitw == 1.25) {
       return 'kiso';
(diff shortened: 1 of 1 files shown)


yanfali:
@noroadsleft any negative consequences on weird layouts?

noroadsleft:
> @noroadsleft any negative consequences on weird layouts?

One minor regression: Keys of `less than 1u height` and `width > height` and are now rendered with 1u height.

[handwired/videowriter](https://config.qmk.fm/#/handwired/videowriter/LAYOUT), top row (Current | This Branch):

image


[converter/palm_usb/stowaway](https://config.qmk.fm/#/converter/palm_usb/stowaway/LAYOUT), right-hand edge (Current | This Branch):

image


Keys of `less than 1u height` and `height > width` are currently fine.

fauxpark:
I can kinda fix that with `unith >= unitw || unith !== 1`, but then the 1u keys become `custom` for some reason.

fauxpark:
Okay, this seems to work now.

noroadsleft:
Thanks!

PR #779, opened 2020-07-28
Add metamechs/timberwolf default keymap

Croktopus:


Croktopus:
thanks for the quick merge. any recommendations on adjusting the formatting, like is there a template or should i just try to be a bit more exacting when i copy an existing file

PR #795, opened 2020-08-22
Add Moonlander default keymap

drashna:
I believe this is how you do this.


drashna:
To bo honest, I got kind of lazy with the default keymap. It's what we have in our fork, however, it's not what the Oryx configurator uses as the default.  

I should/need to fix the default keymap.  However, the json is more closely aligned with what Oryx outputs, and would prefer to use that.

As for the webUSB key, it's mapped to KC_NO, actually, since webUSB is never enabled in the QMK repo.

public/keymaps/m/moonlander_default.json
```diff
@@ -227,5 +227,4 @@
             "KC_WBAK"
         ]
     ],
```
noroadsleft, line 229:
```suggestion
    ]
```

noroadsleft:
Thanks!

PR #811, opened 2020-09-08
remove unuse information

marksard:
When I checked, I didn't see this added keymap in the QMK Configurator, so I'll remove the unnecessary parts.
Regards,

### DIFF
--- a/public/keymaps/m/marksard_rhymestone_rev1.json
+++ b/public/keymaps/m/marksard_rhymestone_rev1.json
@@ -1,7 +1,4 @@
 {
-    "version": 1,
-    "notes": "",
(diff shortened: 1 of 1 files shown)


noroadsleft (approved):
# What Happened Here:

When you export a keymap JSON file from QMK Configurator, the export contains `version`, `notes`, `documentation` and `author` keys. Author and Notes are only filled in if the user uses the Print Keymap functionality (these can be set there so they are on the printed page). `documentation` aims to instruct the user how to use the export. `version` is internal, and is basically "What version of QMK Configurator keymap JSON export is this?" Currently, there is only one defined version.

All this is to say: QMK Configurator exports more data into keymap JSON files than is required to be provided when one wants to *import* a keymap JSON.

PR #814, opened 2020-09-11
Fixed rename marksard keymaps

marksard:
I forgot to add _default to the .json file name, so I changed it. I'm very sorry.


noroadsleft (approved):
Well I missed this during the PR reviews so this is my fault too.

Good catch!

PR #815, opened 2020-09-15
adjust json file to keymap.c file

greenjack-z:


noroadsleft:
This doesn't match the default keymap that's currently on `qmk_firmware` repository - is that deliberate?

greenjack-z:
> This doesn't match the default keymap that's currently on `qmk_firmware` repository - is that deliberate?

Yes, it if final at the moment and will stay as a default I believe. Anyway it's definitely better than it is now.

noroadsleft:
> Yes, it if final at the moment and will stay as a default I believe. Anyway it's definitely better than it is now.

So is the plan that the default keymap on `qmk_firmware` is going to be updated to this?

greenjack-z:
> So is the plan that the default keymap on `qmk_firmware` is going to be updated to this?
Yes, I will do it as a default.

noroadsleft (approved):

noroadsleft:
Thanks!

PR #836, opened 2020-10-28
Add Configurator Keyset Theme "GMK Fro.Yo" 

The-Royal:
# [Fro.Yo Theme](https://i.imgur.com/jMP5zCs.jpg)

## New Files:

- [x] (1)`src/components/colorways/gmk/fro_dot_yo.js`
   - Add key color overrides

## Modified Files:

- [x] (1) `src/components/colorways/gmk/index.js`
   - New file path reference

- [x] (1) `src/scss/colorways.scss`
   - Define key colors

- [x] (1) `src/scss/pantone.scss`
   - Add x4 new PANTONE swatches

- [ ] (1) `src/components/Main.vue`
   - Add filter to insert "." for menu item ***(No idea if this works)***

### DIFF
--- a/src/components/Main.vue
+++ b/src/components/Main.vue
@@ -120,7 +120,8 @@ export default {
           .replace(/Kat/, 'KAT')
           .replace(/Wob/, 'WOB')
           .replace(/Ta/, 'TA')
-          .replace(/ ?Plus/g, '+');
+          .replace(/ ?Plus/g, '+')
+          .replace(/ ?Dot/g, '.');
       });
     },
     redditPost() {
--- a/src/components/Main.vue
+++ b/src/components/Main.vue
@@ -121,7 +121,7 @@ export default {
           .replace(/Wob/, 'WOB')
           .replace(/Ta/, 'TA')
           .replace(/ ?Plus/g, '+')
-          .replace(/ ?Dot/g, '.');
+          .replace(/ ?Dot ?/g, '\.');
       });
     },
     redditPost() {
--- a/src/scss/colorways.scss
+++ b/src/scss/colorways.scss
@@ -575,7 +575,8 @@
   // placeholder
 }
 
-.gmk-fro-dot-yo-key, .gmk-fro-dot-yo-mod {
+.gmk-fro-dot-yo-key,
+.gmk-fro-dot-yo-mod {
   background: $color-gmk-abs-L9;
   color: $color-gmk-abs-2B;
 }
--- a/src/scss/colorways.scss
+++ b/src/scss/colorways.scss
@@ -575,7 +575,7 @@
   // placeholder
 }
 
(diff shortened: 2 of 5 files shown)


noroadsleft (changes requested):

src/components/Main.vue
```diff
@@ -120,7 +120,8 @@ export default {
           .replace(/Kat/, 'KAT')
           .replace(/Wob/, 'WOB')
           .replace(/Ta/, 'TA')
-          .replace(/ ?Plus/g, '+');
+          .replace(/ ?Plus/g, '+')
+          .replace(/ ?Dot/g, '.');
```
noroadsleft, line 124: I want to see if we can work out a better solution for this. For reasons I have yet to work out, attempting to capture the space after `Dot` prevents the UI from fully loading. I'm starting to wonder if it wouldn't be better to just assign the keyset name as a string in the `.js` file and load it directly, instead of manipulating it via regexes.

@yanfali Thoughts?

The-Royal: There's no way for me to play around with this is there? Not easily anyway?

yanfali: It's probably because it's a regex and '.' is actually a match all character. You might look at some unicode replacements that don't trigger the regex.

The-Royal: `.` ?

The-Royal: Normal Period  -`Fro.Yo`

"One Dot Leader" -`Fro․Yo`

noroadsleft: Today I learned that Firefox, Chrome and Brave all will still parse escaped characters in a `.replace` regex. yanfali's comment about `.` being the regex match all character prompted me to try escaping it, and it works.

```suggestion
          .replace(/ ?Dot ?/g, '\.');
```

noroadsleft (changes requested):

src/scss/colorways.scss
```diff
@@ -575,6 +575,34 @@
   // placeholder
 }
 
+.gmk-fro-dot-yo-key, .gmk-fro-dot-yo-mod {
```
noroadsleft, line 578: Apparently the CI doesn't like this statement as one line.

```suggestion
.gmk-fro-dot-yo-key,
.gmk-fro-dot-yo-mod {
```

noroadsleft (changes requested):

noroadsleft:
Also, I don't like how low contrast most of the keys are (WCAG Contrast ratings are all below 2.0 except for one). I may want to color shift this set later just so it's more readable.

noroadsleft (approved):
I still want to improve the contrast of this theme, but this will do for now.

noroadsleft:
Thanks!

PR #868, opened 2021-01-05
ci: switch to github actions

zekth:
Basically adds 2 workflows in parallel of travis for the transient state.
The build process will publish the compiled website to `test-gh-pages` branch, which is harmless for the project. Once validated, can switch to `gh-pages`.


.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: cypress-io/github-action@v2
+        with:
+          start: yarn serve
+          wait-on: 'http://localhost:8080'
+      - name: build
+        run: |
+          echo VUE_APP_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
+          yarn run build
+          echo config.qmk.fm > dist/CNAME
+      - name: Deploy
```
zvecr, line 42:
```suggestion
      - name: Deploy
        if: ${{ github.event_name != 'pull_request' }}
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
+name: Website build
+
+on:
+  push:
+    branches:
+      - master
+
```
zvecr, line 7:
```suggestion
on:
  pull_request:

```

.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          deploy_key: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./dist
+          publish_branch: test-gh-pages
+          force_orphan: true
+          user_name: 'QMK Bot'
+          user_email: 'hello@qmk.fm'
+          commit_message: 'Deploy to GitHub Pages'
```
zvecr, line 51: We currently use the following action everywhere else, so it would be good to align if possible.

Something like...
```suggestion
      uses: JamesIves/github-pages-deploy-action@3.7.1
      with:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        BRANCH: test-gh-pages
        FOLDER: ./dist
        GIT_CONFIG_NAME : QMK Bot
        GIT_CONFIG_EMAIL: hello@qmk.fm
```

.github/workflows/pr.yml
```diff
@@ -0,0 +1,38 @@
+name: Pull Request Test
```
zvecr, line 1: With the changes above, this file can be removed.

zvecr (changes requested):
Maybe something like the suggestions to have a single file, and reduce the duplication?

.github/workflows/build.yml
zekth: This really depends on how you work regarding the difference between build and pr. doing the change

zvecr (approved):

PR #878, opened 2021-02-01
New Breeze Rev1 default

eithanshavit:
## Description
Adding a new default keymap for Breeze Rev1.
Also, matching the Rev0 keymap with the commit hash state.

### DIFF
--- a/public/keymaps/a/afternoonlabs_breeze_rev1_default.json
+++ b/public/keymaps/a/afternoonlabs_breeze_rev1_default.json
@@ -2,7 +2,7 @@
    "keyboard": "afternoonlabs/breeze/rev1", 
    "keymap": "default", 
    "layout": "LAYOUT", 
-   "commit": "b32b928ff3dea8a3d396ff44d491bb3d4a95a2c8",
+   "commit": "64da7413a0c2b3ea555288f5d296e3a98a0da6b4",
    "layers": [
       [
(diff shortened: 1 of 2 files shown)


public/keymaps/a/afternoonlabs_breeze_rev1_default.json
```diff
@@ -0,0 +1,29 @@
+{
+   "keyboard": "afternoonlabs/breeze/rev1", 
+   "keymap": "default", 
+   "layout": "LAYOUT", 
+   "commit": "b32b928ff3dea8a3d396ff44d491bb3d4a95a2c8",
```
noroadsleft, line 5: This change should also go on the rev0 keymap.

```suggestion
   "commit": "64da7413a0c2b3ea555288f5d296e3a98a0da6b4",
```

noroadsleft (changes requested):
I also note that you have the `LOWER`/`MO(1)` and `RAISE`/`MO(2)` keycodes in opposite order from how they are on `qmk/qmk_firmware`, but the `qmk_firmware` keymap should be updated to match this behaviour IMO, because the `RAISE` keycode actually blocks access to `KC_P0` on that side anyway.

eithanshavit:
Thanks @noroadsleft, I've committed the new hash. 
Also, thanks for the catch on the MO order, it makes sense to update qmk_firmware. I'll do it in a separate PR.

eithanshavit:
Updated `qmk_firmware` to match in https://github.com/qmk/qmk_firmware/pull/11779

noroadsleft (approved):

noroadsleft:
Thanks!

PR #881, opened 2021-02-05
Unnecessary quoting & escaping of keymap JSON

liyang:
## Description

Upon every keymap JSON upload/edit/download cycle, my email address in the
`author` field gains another pair of `&amp;`. Likewise for `notes`.

We shouldn't HTML-entity `encode` here, as (1) it's not appropriate for
JSON, and (2) JSON serialisation will already escape anything that needs it.

FWIW, when I upload a pristeen keymap JSON and click ‘Print Keymap’, the `author` and `notes` textarea/input shows escaped HTML entities, so it's not necessary to escape for those either.

Bonus removal of extraneous `"`s in the `documentation` field of the keymap template.

### DIFF
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -127,7 +127,6 @@ import Vue from 'vue';
 import { mapMutations, mapActions, mapState, mapGetters } from 'vuex';
 import first from 'lodash/first';
 import isUndefined from 'lodash/isUndefined';
-import escape from 'lodash/escape';
 import { clearKeymapTemplate } from '@/common.js';
 import { PREVIEW_LABEL } from '@/store/modules/constants';
 import {
@@ -350,8 +349,8 @@ export default {
 
       if (!isUndefined(data.author)) {
         const { author, notes } = data;
-        this.setAuthor(escape(author));
-        this.setNotes(escape(notes));
+        this.setAuthor(author);
+        this.setNotes(notes);
       }
 
       // remap old json files to new mappings if they need it
--- a/src/store/modules/keymap.js
+++ b/src/store/modules/keymap.js
@@ -30,7 +30,7 @@ const state = {
     keymap: {
       version: 1,
       documentation:
-        '"This file is a QMK Configurator export. You can import this at <https://config.qmk.fm>. It can also be used directly with QMK\'s source code.\n\nTo setup your QMK environment check out the tutorial: <https://docs.qmk.fm/#/newbs>\n\nYou can convert this file to a keymap.c using this command: `qmk json2c {keymap}`\n\nYou can compile this keymap using this command: `qmk compile {keymap}`"\n'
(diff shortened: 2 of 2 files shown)


yanfali:
OK, I think I've never noticed this because lodash works differently in chrome and defers to the native javascript escape. You should never trust user input, so I would suggest a different patch. Remove the import from 'escape' from lodash, and leave the escapes. The native escape leaves @ alone.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape

This should then work correctly. p.s. when reporting bugs, repeatability is important, so please include the browser and the OS so we can reproduce your symptoms. Also make sure to test in an incognito window to eliminate plugins.

Thanks for the report.

src/components/ControllerBottom.vue
```diff
@@ -350,8 +349,8 @@ export default {
 
       if (!isUndefined(data.author)) {
         const { author, notes } = data;
-        this.setAuthor(escape(author));
```
yanfali, line 353: leave these in and use the global escape.

liyang:
Sorry I should have included exact steps for reproducing the problem. Save the following `keymap.json`:
```
{
  "author": "Me <myself@and-i.com>",
  "keyboard": "xd60/rev3",
  "keymap": "liyang",
  "layers": [],
  "layout": "LAYOUT_all",
  "notes": "# Liyang's ISO Keymap for XIUDI's 60% XD60 (rev3) PCB",
  "version": 1
}
```
and import it. On `Home.vue`, neither `author` nor `notes` are shown, so on the surface everything seems fine. Now click “PRINT KEYMAP” / navigate to `Print.vue` where the aforementioned _are_ visible. You'll see the following:
qmk-configurator

Exporting that keymap gives the following:
```
{
  "version": 1,
  "notes": "# Liyang&#39;s ISO Keymap for XIUDI&#39;s 60% XD60 (rev3) PCB",
  "documentation": "\"This file is a QMK Configurator export. […]\"\n",
  "keyboard": "xd60/rev3",
  "keymap": "liyang",
  "layout": "LAYOUT_all",
  "layers": [],
  "author": "Me &lt;myself@and-i.com&gt;"
}
```

So `Me <myself@and-i.com>` is being turned into
```
Me &lt;myself@and-i.com&gt;
Me &amp;lt;myself@and-i.com&amp;gt;
```
ad infinitum each time I import/edit/export a keymap JSON.

> […] The native escape leaves @ alone.
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape
> This should then work correctly.

This wasn't the issue—`@` doesn't even need escaping in either HTML, or as part of a URL. From the above documentation:

>> This function was used mostly for URL queries (the part of a URL following ?)—not for escaping ordinary String literals

So using the global `escape` would produce a slightly different problem. Instead of the above, you get:
```
Me%20%3Cmyself@and-i.com%3E
Me%2520%253Cmyself@and-i.com%253E
```
with every import/edit/export cycle.

----

As an aside:

> You should never trust user input, so I would suggest a different patch.

By the time that `escape` is called, the user-uploaded JSON has already been deserialised, and you're just dealing with a plain JS string. Guarding against malicious user input has already been taken care of by the JSON parser. Indeed if you try to upload a malformed JSON file, you rightly get the error “Sorry, that doesn't appear to be a valid QMK keymap file.”

yanfali:
I never really anticipated your use case, but it does seem reasonable. I'm somewhat paranoid so that's the reason the escaping was done in the first place, but I think you are probably correct JSON.parse should prevent anything nefarious from happening.

yanfali (approved):
Thanks for your contribution.

PR #894, opened 2021-03-06
Add Ctrl+Shift modifier key

AndrewKvalheim:
I didn't see a way to create a [<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>U</kbd>](https://help.gnome.org/users/gnome-help/stable/tips-specialchars.html.en#ctrlshiftu) key using the configurator. [`RCS(kc)`](https://docs.qmk.fm/#/feature_advanced_keycodes?id=modifier-keys) would work for me. Can we add that?

### DIFF
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -282,6 +282,12 @@ export default [
     type: 'container',
     title: 'LCTL + LALT'
   },
+  {
+    name: 'RCS',
+    code: 'RCS(kc)',
+    type: 'container',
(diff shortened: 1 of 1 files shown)


yanfali:
Have you tried `ANY(C(S(U)))` ?

AndrewKvalheim:
Ah, somehow I missed Any. That's good enough for me, thanks.

yanfali:
In principle, I'm not against adding it, it just adds a bit to the overall clutter of the UI from less common use cases.

PR #912, opened 2021-04-21
Add default keymaps for CozyKeys Bloomer v2 and v3

pcewing:
## Description

I am in the process of adding the CozyKeys Bloomer keyboard to QMK and this PR is to add the default keymaps to the configurator.

The corresponding PR in the QMK repository is:
https://github.com/qmk/qmk_firmware/pull/12639

**Do not merge yet.** This is intentionally set up as a draft PR for two reasons:
- The commit hash for the keyboard doesn't exist until the corresponding PR in the QMK repository is merged, at which point I will update the `"commit"` fields here
- I have not been able to test importing of these files
    - I was able to load my local `info.json` into the configurator, set up the keymap, and export it; however, it looks like importing a keymap doesn't work when a local `info.json` file is being used
    - My understanding is that the configurator parses for `info.json` files in the `master` branch of QMK, so I will test this as soon as the corresponding PR is merged in the QMK repo

### DIFF
--- a/public/keymaps/c/cozykeys_bloomer_v2_default.json
+++ b/public/keymaps/c/cozykeys_bloomer_v2_default.json
@@ -13,12 +13,12 @@
       "KC_LCTL", "KC_LGUI", "KC_LALT", "KC_GRV",  "MO(1)",   "KC_BSPC", "KC_DEL",   "KC_DOWN", "KC_ENT",  "KC_SPC",  "KC_LBRC", "KC_RBRC", "KC_RALT", "KC_RGUI", "KC_RCTL"
     ],
     [
-      "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",    "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "KC_NO",   "RESET",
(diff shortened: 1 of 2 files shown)


noroadsleft:
>       * I was able to load my local `info.json` into the configurator, set up the keymap, and export it; however, it looks like importing a keymap doesn't work when a local `info.json` file is being used

You can work around this by doing the following:

- Copy the keymap JSON file you wish to test to `"public/keymaps/p/Preview info.json_default.json"` (note the quotes and space!)
  ```shell
  cp public/keymaps/c/cozykeys_bloomer_v2_default.json "public/keymaps/p/Preview info.json_default.json"
  ```
- Load the `info.json` file which corresponds to the keymap you copied using <kbd><kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd></kbd>
- Click the `Load Default` button

image

pcewing:
Ah, I wasn't running the configurator locally but looks like it should be easy to. Also, I already see an issue in that layout (Win key where N should be).

Will fix it up and test. Thanks!

pcewing:
The PR in the main QMK repository was merged yesterday. I fixed up a few mistakes in the keymaps and tested them so this should be ready now!

noroadsleft (commented):
Not really an error per se, but your FN layers here have `KC_NO` where the `qmk_firmware` keymaps use `KC_TRNS`, so firmware compiled from source will behave differently on the Function layer than Configurator builds will.

If you want to keep them `KC_NO` here, let me know.

pcewing:
Thanks! Didn't notice that, just made that change. :)

noroadsleft (approved):

noroadsleft:
Thanks!

PR #944, opened 2021-07-04
[Feature] Layer Editor

yanfali:
Iterating through ideas for editing layers within the configurator.


.nvmrc
```diff
@@ -1 +1 @@
-11.12
+14
```
yanfali, line 1: Node 15 breaks node-sass.

yanfali:
The basic concept I'm working through right now is in memory manipulation of the layer data and re-writing of LT/MO references so they are re-written when someone moves around a layer. This should largely happen within the data model with the UI sending commands to the vuex module.

Initial idea is to show a minimal keycode list on each layer ignoring KC_NO/TRNS keycodes as they are empty information in this context.

### Some ideas:
 - compression function to coalesce empty layers and minimize flash space used by keymap (noroadsleft)
 - disable buttons on certain layers (0 and 15 disable move down and up respectively)
 - grey out empty layers (entirely filled with KC_NO/KC_TRNS) in UI
 - probably can't deal with ANY keycodes effectively so ignore them
 - use vue transitions to convey motion between layers  
 - ___warn user about intermediate layers that are filled with KC_NO (layer trapping)___ - stretch
 - ___try to estimate the memory used by a layer(s)___ - stretch
 - ___mini rendering of keymap on hover___ - stretch

PR #947, opened 2021-07-09
Fix printing keymaps with empty layer - Issue 852

### LINKED ISSUE #852: Print Keymap is blank if one of your layers is empty
I have layers 0-1 set up, 2 is blank (everything is N/A), and layers 3-6 are full. When I press print keymap I get a blank screen. I was able to get a keymap to print when I filled layer 3.

I'm purposely leaving 3 empty so that I can put another layout there in the future.

callmetwan:
## Description

Turns out the PrintKeymap component was being rendered incorrectly with an incorrect dynamic binding. There is only one component to render so directly referencing both solves the issue and reduces complexity.

Removes a prettier ignore in PrintKey.

Clarifies a variable name in Print.
 
Fixed #852

### DIFF
--- a/src/components/PrintKey.vue
+++ b/src/components/PrintKey.vue
@@ -5,14 +5,15 @@
     :class="myclasses"
     :style="mystyles"
     :title="displayName"
-  >{{ displayName }}
+  >
+    {{ displayName }}
   </div>
 </template>
 <script>
-import BaseKey from "./BaseKey";
+import BaseKey from './BaseKey';
 
 export default {
-  name: "print-key",
+  name: 'print-key',
   props: {
     layer: Number
   },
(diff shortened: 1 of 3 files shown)


yanfali:
I think your prettier settings may be conflicting. The project uses single quotes in code and double quotes in markup. Other than that looks good. I'll finish reviewing it tomorrow. Thanks

callmetwan:
You were write, I needed to include .vue files in my editors list of files to run prettier against.

yanfali:
@callmetwan you might want to run the server in dev mode and or invoke `yarn prettier:check` or `yarn prettier:fix`. My editor does the prettier fixes on save, so I don't usually see this kind of issue. Thanks

yanfali:
@callmetwan so when I test this it doesn't actually solve the issue. Here's what I did.

Screen Shot 2021-07-09 at 10 00 22 PM

This is a keymap with 3 layers. The middle one is empty and has no data in it. It's effectively not initialized yet. This is by design, but maybe that's the wrong design. If you now go to the print keymap, the code crashes because there's no layer 1 and so the code doesn't work.

I think what you may have fixed was a warning, that was not fatal, just annoying, which in itself is valuable but doesn't fix the underlying bug.

yanfali:
Screen Shot 2021-07-09 at 10 03 01 PM

I'll have to think about what the right way to fix this is. In previous versions we allowed empty layers, but this had the unfortunate side effect of making the hex/bin images larger, which was undesirable if you had a lot of features enabled, so we made the decision ignore layers that have no keycodes.

The case I discovered is actually a corner case where you initialize layer 2 but 1 has never been visited.

callmetwan:
This is weird, I'm not getting that at all and I as far as I can tell I tested the same way. Here is video showing exactly what I did:


https://user-images.githubusercontent.com/526768/125167131-ffd81180-e16c-11eb-8ad8-a5d0af99fef8.mov

As you can see it prints for me as well as the list of keymap indices being accurate.

yanfali:
Ok. I'll double check. Maybe I have caching issues. Thanks

yanfali:
Well that's weird. I guess I was pointing at a cached version. Thanks for verifying that for me.

yanfali:
Thanks so much for your contribution!

callmetwan:
Thanks, I'm glad to do it!!

Just to confirm, this did end up working for you? I want to make sure that there isn't some weird discrepancy between our local systems.

callmetwan:
Also, what is the release process? I'd like to post in the original issue so the reporter can be updated.

yanfali:
Yes I got it to work finally. Thanks

yanfali:
Once merged it is deployed automatically. Issue was closed because it of
reference.

PR #951, opened 2021-07-10
Fix two buttons using same leavePrint ID

callmetwan:
## Description

An ID can only be applied to one element at a time. The buttons within Print.vue both had the same ID. Further, it looks like they were just using it for styling. This styling is already achieved with the `ui-button` class so I have removed the IDs and applied the class instead.


yanfali (approved):
Thanks

PR #959, opened 2021-07-26
New keymap for Sofle rev1 for Flare576

Flare576:
## Description

Uses outside bottom buttons for swapping between "Num Pad" and "Game" layouts, and various LT keys for quick access to mouse and symbols. Also focuses on the `,` and `.` keys for brace characters:

| Combo | `,`/`.` |
|-------|-----|
| Shift | `<`/`>` |
| Mouse | `[`/`]` |
| Shift + Mouse | `{`/`}` |
| Esc | `(`/`)` |

Feel free to visit https://github.com/Flare576/sofle for more information

Ties to https://github.com/qmk/qmk_firmware/pull/13723

### DIFF
--- a/docs/keymaps/s/sofle_rev1_flare576.json
+++ b/docs/keymaps/s/sofle_rev1_flare576.json
@@ -1,0 +1,52 @@
+{
+  "version": 1,
+  "notes": "Instead of up/down, uses layer toggles on ends, incorporates Mouse/Game layers.",
+  "keyboard": "sofle/rev1",
+  "keymap": "Flare576",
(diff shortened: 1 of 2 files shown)


yanfali:
@Flare576 hey. Sincerely thanks for your contribution. Unfortunately we don't currently support more than 1 default keymap per keyboard so this wouldn't really make sense to add to the repo. We hope in a future API rev that it would be possible to retrieve multiple keymaps we keyboard but this is currently an unsolved problem. The way we default keymaps currently is kind of a hack and it could be fixed, but because we have chosen not to at this time as it would be better if this were solved at the API layer.

As a work around, I would use the cloud button and just load it from a raw github location. Thanks for your understanding.

PR #968, opened 2021-08-11
Remove JQuery, technical debt

yanfali:
- Factor away jquery.js, a left over from before the vue port
   - move code into appropriate vuex or files
   - api/util/exclusion_list/remap.js
 - fix npm package dependency issues caused by deps of deps
 - switch from promise style code blocks to async await with appropriate try catch
 - upgrade to node 14


noroadsleft (changes requested):
25a865bc7a87d03831ae8360b123743006838aca causes the readme files to not load anymore.

yanfali:
Good catch.

noroadsleft (approved):

PR #983, opened 2021-08-29
Switch upload and download icons so it makes sense

kleenkanteen:
Look at this:
export button

The export button has a import icon. Looks odd. I changed my layout multiple times as I've improved it and this gets me every time.

Switched the font awesome icons for the 2 icons, download and upload.

### DIFF
--- a/src/components/ControllerBottom.vue
+++ b/src/components/ControllerBottom.vue
@@ -25,7 +25,7 @@
         @click="exportJSON"
         v-tooltip.bottom="$t('downloadJSON.title')"
       >
-        <font-awesome-icon icon="download" size="lg" fixed-width />
+        <font-awesome-icon icon="upload" size="lg" fixed-width />
       </button>
       <span class="label-button">{{ $t('downloadJSON.label') }}</span>
       <button
@@ -33,7 +33,7 @@
         v-tooltip.bottom="$t('importJSON.title')"
         @click="importKeymap"
       >
-        <font-awesome-icon icon="upload" size="lg" fixed-width />
+        <font-awesome-icon icon="download" size="lg" fixed-width />
       </button>
       <button
         id="import-url"


skullydazed:
Maybe it's the tooltip that needs to change? Would Download/Upload or Save/Load work better instead of Export/Import?

From an iconography standpoint ISTM that the rectangle with two dot represents the user's computer, and the arrow indicates which direction the JSON will move in. From that perspective swapping them would be confusing.

skullydazed:
Another thought- maybe instead they should swap sides of `keymap.json`. With the Full Source and Firmware download buttons being on the far right there's a certain consistency in making the "Download keymap.json" action be on the right too.

kleenkanteen:
Hmm, I was thinking export is you are bringing it out of the site, so export the layout out, hence the arrow going up. Import is going into the site, so arrow in. So the website is the point of reference.

For the Firmware download buttons, I think it's fine as it's not deceptive at the least. But you could if you want you could, I can see why it makes more sense.

skullydazed:
> For the Firmware download buttons, I think it's fine as it's not deceptive at the least. But you could if you want you could, I can see why it makes more sense.

To clarify- I'm not proposing changing the current firmware download buttons. Only the current import/export order.

noroadsleft:
> From an iconography standpoint ISTM that the rectangle with two dot represents the user's computer, and the arrow indicates which direction the JSON will move in. From that perspective swapping them would be confusing.

This is exactly what it's representing. Export is Download; Import is Upload.

noroadsleft:
I'm personally hesitant to move the buttons around or switch the icons; they've been this way since we launched this version of QMK Configurator in February of 2019, and "muscle memory" is a thing. I'd be much more receptive to changing the wording, personally.

kleenkanteen:
Would you accept a vote from people who used it a few to couple times and are new like me?

noroadsleft:
I understand your issue, and appreciate that you've actually gone and endeavored to fix it, but I'm not convinced this is actually an improvement. These buttons have been this way since we launched this version of Configurator two and a half years ago (and there was a publicly-accessible beta period before that), and you're the first user to complain about this in that time.

I actually find your change more confusing, because the icon swap make it look like the button functions have moved when they haven't - the left is still Export and the right is still Import.

kleenkanteen:
I may just be the first one bothered enough to complain about it :p 

I think it'd make more sense if the icons are switched because I see the website as the point of reference.

Can I do a poll for just 1 month for new users and see what they think?

yanfali:
Sorry but this is just bike shedding.

kleenkanteen:
How is it bike shedding? And how is any UI improvement too small to be considered? Wheres the point where something is trivial or not?

zvecr:
image

> I actually find your change more confusing

+1 for this. Improving usability is a fine goal, but this change isnt it.

yanfali:
> How is it bike shedding? And how is any UI improvement too small to be considered? Wheres the point where something is trivial or not?

You've failed to convince 4 people who are collaborators that your change merits attention. I would move on, this is subjective and we as a group don't agree with your opinion.

PR #998, opened 2021-09-29
fixed incorrect display while pressing F13

yulei:
The displayed result was not correct while pressed F13 under keyboard tester mode.

(C0K5PP3VFWM1`0J1_ZS%LT

## Description
it should be:

U8K` 8`3T8_HE$9J{GDWURW

### DIFF
--- a/src/store/modules/tester/codeToPos.js
+++ b/src/store/modules/tester/codeToPos.js
@@ -12,7 +12,7 @@ const ROW1 = [
   'F10',
   'F11',
   'F12',
-  'F13',
+  'PrintScreen',
   'ScrollLock',
   'Pause'
 ];


yanfali:
Hey, thanks for the PR. Did you see my comment about macOS? I think this might just be a macOS quirk.

yulei:
The F13 -> PSCR convert was made by the Tester, it's not the OS's behaviour.

yanfali:
> The F13 -> PSCR convert was made by the Tester, it's not the OS's behaviour.

So what does it do on windows? I'm testing on MacOS and now it prints nothing. I'm wondering if we need to make this OS dependent. There is a way to do that using navigator. That's how we detect mac keys for icons.

yanfali:
OK. I guess this is just a macOS quirk, and we should just fix it. Thanks

yulei:
> > The F13 -> PSCR convert was made by the Tester, it's not the OS's behaviour.
> 
> So what does it do on windows? I'm testing on MacOS and now it prints nothing. I'm wondering if we need to make this OS dependent. There is a way to do that using navigator. That's how we detect mac keys for icons.

The behavior was same as the MacOS of the current version.  This was not OS dependent, it was caused by the Tester's codes.
It print nothing while using the PR because there are no keycode's name for F13 currently.

PR #1013, opened 2021-11-04
Update handwired_split89_default.json

jurassic73:
Left control key map updated to use control instead of shift.   (DOH!)


jurassic73:
One last fix and my configs should be good.  Sorry for multiple fixes.

noroadsleft (commented):
Have you tried flashing this to your keyboard? Comparing this to the default `keymap.c` in qmk/qmk_firmware, it looks like they don't match.

jurassic73:
@noroadsleft - I've verified it works.  I can currently use QMK configurator to create and download working firmware that works fine minus my left control key working as a shift key. I just had to reassign left ctrl to the left ctrl key(this pull request). I performed this test last night and again just now from the main QMK page to load this firmware on my keyboard and, minus the one mismapped key, it works great.

https://config.qmk.fm/#/handwired/split89/LAYOUT

This pull request should get this in full working order by default.  Thanks for verifying. :)

noroadsleft:
Then my guess is your default firmware on qmk/qmk_firmware won't assign the keys correctly.

jurassic73:
it is currently though?

noroadsleft:
I mean https://github.com/qmk/qmk_firmware/blob/45b35e6ea13d8c184d9cbd23e4bd3868e66d2933/keyboards/handwired/split89/keymaps/default/keymap.c#L23-L42 probably doesn't work correctly.

QMK Configurator keymaps need to assign their keycodes in the same order that `keymap.c` uses – this keymap doesn't, which means one or the other is wrong.

noroadsleft (approved):

noroadsleft:
Thanks!

jurassic73:
@noroadsleft I appreciate you looking at this and your time!  This works now.  Forgive my ignorance if this should be written out differently.  This keymap to wiring matrix was really confusing for me.  Thanks again.

PR #1021, opened 2021-11-23
Stop service fonts via Google Fonts

WhyNotHugo:
Switch from serving fonts from Google Fonts to bundling them with the
app at build time.

This has a few consequences:

- Improved privacy for users: no leaking to Google details about people
  visiting configurator.
- Google Fonts are not versioned and may be updated at any time, which
  can result in breakage for users.
- Google Fonts vary at any time, so improved security like SRI or
  cross-domain hardening are not possible without breakage.
- This makes configurator load properly offline once built, which is
  super handy for those with spotty or metered connections.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,9 @@
     "test:cypress:ci": "start-server-and-test serve http://localhost:8080 test:cypress"
   },
   "dependencies": {
+    "@fontsource/montserrat": "^4.5.1",
+    "@fontsource/roboto": "^4.5.1",
+    "@fontsource/roboto-mono": "^4.5.0",
     "@fortawesome/fontawesome-svg-core": "^1.2.12",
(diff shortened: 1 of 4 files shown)


yanfali (commented):
Seems like a good idea. What licenses? Any distribution issues if built into docker?

WhyNotHugo:
Montserrat is https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL

Roboto is https://www.apache.org/licenses/LICENSE-2.0.html

WhyNotHugo:
Fixed prettier issues.

WhyNotHugo:
Nice, prettier is happy now :relieved:

yanfali (approved):
Thank you

yanfali:
Looks like it's working well on config.qmk.fm. Thanks again

PR #1027, opened 2021-12-13
Add Splitography to the steno keyboards

jeandeaual:
## Description

The [Splitography steno keyboard](https://softhruf.love/products/soft-hruf-erl) was [recently added to QMK](https://github.com/qmk/qmk_firmware/pull/15133).
This PR simply adds it to the list of steno keyboards.

### DIFF
--- a/src/store/modules/app/actions.js
+++ b/src/store/modules/app/actions.js
@@ -9,7 +9,7 @@ import {
 import { getPreferredLayout, getExclusionList } from '@/util';
 import { localStorageSet, CONSTS } from '@/store/localStorage';
 
-const steno_keyboards = ['gergo', 'georgi'];
+const steno_keyboards = ['gergo', 'georgi', 'splitography'];
 
 const actions = {
   /**


jeandeaual:
I also tried [adding a default layout](https://github.com/jeandeaual/qmk_configurator/commit/14415589148aeffed32dd56a4a8e2d8d1c7ac04b) for the Splitography, but it seems that it can't compile since the generated C file doesn't include `keymap_steno.h`.

noroadsleft:
Configurator can't compile the Georgi either; I'm thinking Configurator needs another endpoint to tell the compiler to include the `keymap_steno.h` before this will work.

@yanfali any ideas?

yanfali:
> Configurator can't compile the Georgi either; I'm thinking Configurator needs another endpoint to tell the compiler to include the `keymap_steno.h` before this will work.
> 
> @yanfali any ideas?

Not if it's on the API side. We have limited control over headers.

skullydazed:
> I'm thinking Configurator needs another endpoint to tell the compiler to include the `keymap_steno.h` before this will work.

Someone needs to extend the keymap.json format to support steno, then the QMK cli can generate the `keymap_steno.h` file during compilation.

PR #1028, opened 2021-12-14
Fix Portuguese to use the correct endonym Português

peepeetee:

### DIFF
--- a/src/store/modules/app/state.js
+++ b/src/store/modules/app/state.js
@@ -69,7 +69,7 @@ const state = {
     { value: 'es', label: 'Español' },
     { value: 'ja', label: 'Japanese' },
     { value: 'ru', label: 'Русский' },
-    { value: 'pt-BR', label: 'Português' },
+    { value: 'pt-BR', label: 'Português brasileiro' },
     { value: 'pl-PL', label: 'Polski' },
     { value: 'ms', label: 'Bahasa Malaysia' },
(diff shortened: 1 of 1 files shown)


src/store/modules/app/state.js
```diff
@@ -69,7 +69,7 @@ const state = {
     { value: 'es', label: 'Español' },
     { value: 'ja', label: 'Japanese' },
     { value: 'ru', label: 'Русский' },
-    { value: 'pt-BR', label: 'Brasileira' },
+    { value: 'pt-BR', label: 'Português' },
```
fauxpark, line 72:
```suggestion
    { value: 'pt-BR', label: 'Português (Brasil)' },
```

fauxpark (changes requested):

peepeetee:
@fauxpark The docs just says Português, we should try to keep them synced somehow. Should I go and make the change over there too?
image

noroadsleft (approved):

noroadsleft:
fauxpark can feel free to disagree with me here, but I don't see any reason that the docs and Configurator have to use the same strings for language names. It's also been quite some time since I worked on docs translations, but I seem to recall that the menu you screen-captured is width-limited in the docs theme we use, and so longer strings get cut off visually anyway.

I'm good with the state of this PR as-is; merging. Thanks!

PR #543, opened 2019-10-17
Added Japanese translation

shelaf:

PR #1029, opened 2021-12-14
Update Japanese label to use endonym, reorder languages according to google's ordering

peepeetee:

### DIFF
--- a/src/store/modules/app/state.js
+++ b/src/store/modules/app/state.js
@@ -65,13 +65,13 @@ const state = {
   languages: [
     { value: 'en', label: 'English' },
     { value: 'de', label: 'Deutsch' },
-    { value: 'fr', label: 'Français' },
     { value: 'es', label: 'Español' },
-    { value: 'ja', label: '日本語' },
-    { value: 'ru', label: 'Русский' },
-    { value: 'pt-BR', label: 'Português brasileiro' },
-    { value: 'pl-PL', label: 'Polski' },
+    { value: 'fr', label: 'Français' },
     { value: 'ms', label: 'Bahasa Malaysia' },
+    { value: 'pl-PL', label: 'Polski' },
+    { value: 'pt-BR', label: 'Português brasileiro' },
+    { value: 'ru', label: 'Русский' },
+    { value: 'ja', label: '日本語' },
     { value: 'zh-CN', label: '简体中文' }
   ],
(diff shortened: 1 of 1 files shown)


fauxpark (approved):

mtei:
Thank you for your PullRequest.

However, I don't think this needs to be changed. Because it is the result of intentionally selecting "Japanes".

see https://github.com/qmk/qmk_configurator/pull/543#discussion_r336758422

Tag: @shelaf

fauxpark:
The general convention I see is for language selections to use their own spelling/character set.

The second point mentioned in that comment is nonsensical, pretty much any device which is able to use the Configurator effectively would have the basic CJK characters in at least one font.

shelaf:
It's what I intended, but it's okay to change it as @fauxpark says. However, I think the display order should be changed. Currently in "value" order, but should be in "label" order.

peepeetee:
> It's what I intended, but it's okay to change it as @fauxpark says. However, I think the display order should be changed. Currently in "value" order, but should be in "label" order.

I'm thinking of putting CJKV at the bottom and using alphabetic ordering for Latin/Cyrillic scripts, and stroke order for CJKV. 

Any better ideas?

Also, should I include the reordering in this PR, or open another PR for that?

peepeetee:
My opinion is that sorting by language code is better than sorting by alphabetical/stroke order, as sorting by language code is unambiguous. 
If we are to sort alphabetical, it would introduce interesting dilemmas: should we sort Bahasa Malaysia under M or B? Русский under r, or put it at the end of the list? ~~we can also sort 日本語 under n for maximum confusion~~

shelaf:
IMO, Unicode character code order is good to avoid the complexity.
cf. Google account language settings:
https://myaccount.google.com/language?continue=https%3A%2F%2Fmyaccount.google.com%2Fpersonal-info

The changes can be included in this PR.

peepeetee:
> IMO, Unicode character code order is good to avoid the complexity. cf. Google account language settings: https://myaccount.google.com/language?continue=https%3A%2F%2Fmyaccount.google.com%2Fpersonal-info
> 
> The changes can be included in this PR.

Done

Anything else I should do before this PR could be merged?

noroadsleft (approved):

noroadsleft:
Thanks everyone!

PR #1035, opened 2021-12-23
added keymap for k34

wongjingping:
Adding keymap for k34

## Description

Adding keymap json for k34. Tested on the configurator and it loads.


noroadsleft (changes requested):
1. The file needs to be renamed to `k34_default.json`:
    `git mv public/keymaps/k/k34.json public/keymaps/k/k34_default.json`
2. Why are there eight layers here when the default keymap in the `qmk_firmware` repository only has three?

wongjingping:
> 1. The file needs to be renamed to `k34_default.json`:
>    `git mv public/keymaps/k/k34.json public/keymaps/k/k34_default.json`
Done
> 2. Why are there eight layers here when the default keymap in the `qmk_firmware` repository only has three?
Was reusing another keymap. Changed back to 3 for consistency.

noroadsleft (approved):

noroadsleft:
Thanks!

PR #1037, opened 2021-12-30
Add playlist to QMK Tutorial wizard

redxtech:
## Description
Currently, when you enable the wizard, it brings up a single video. While this is useful, I think it might be useful for users to see that there's a playlist of other related tutorials as well, so I have edited the embed link to include the playlist.


yanfali (approved):
Looks great. Thanks

PR #1044, opened 2022-01-07
add 3 layout's default keymap to lefishe

peepeetee:

### DIFF
--- a/public/keymaps/l/lefishe_wk_sbs.json
+++ b/public/keymaps/l/lefishe_wk_sbs.json
@@ -1,0 +1,22 @@
+{
+  "keyboard": "lefishe",
+  "keymap": "default",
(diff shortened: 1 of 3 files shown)


lyso1 (approved):

noroadsleft (commented):
This won't work.

https://github.com/qmk/qmk_configurator/blob/6ef65962eb09e0841afdfafd74ee555d1e831fc9/src/store/modules/app/actions.js#L30-L44

The `Load Default` button is hardcoded to load `<keyboard_name>_default.json` when clicked. Any keymaps with another filename are completely invisible to QMK Configurator.

I wouldn't say it's a `wontfix`, but I would say "Don't hold your breath."

yanfali:
What is this even. Closing

PR #1049, opened 2022-01-16
Vue-cli -> Vite Build Improvement

yanfali:
Thinking about Macros.
Move build system to Vite as an experiment.

## Description
We are currently using vue-cli for building, this has been a great tool for us, but it's no longer the best option, especially if we want to move to vue3. Improvements in browsers allow them to load assets directly using ES6 imports, so it makes sense to use vite which is

 - a creation of the creator of vue, Evan Yu, so has first class support
 - based on es6 loaders and is extremely fast which means a better developer experience
 - is a path to vue 3 because all the examples you will find these days are using vite
 - allows easier integration of newer technologies which could benefit the project like windicss
 - have I mentioned it's fast?

### Changes

 - transition away from vue-cli
 - transition towards vite
 - fix up places where vite and vue-cli differ, import naming mostly
 - fix verification of keymap json file
   - use a Promise and IIFE to wait for all files to be tested and return a value to shell so scripting can detect an error. Previously it was printing a warning and silently ignoring it.
 - fix up cypress
   - upgrade vue-router and cypress to more recent versions because old code is hard to maintain
   - use better vue-router error handling
     - a source of errors were exceptions being thrown by router to do with Cancelled route updates
     - ignore them explicitly but throw all other errors
   - remove deprecated cypress calls and replace with modern equivalents
   - adjust timings for some tests


yanfali:
@zekth any thing look sus?

.github/workflows/deploy_develop.yml
```diff
@@ -42,13 +42,13 @@ jobs:
...
-          wait-on: "http://localhost:8080"
+          start: yarn dev
+          wait-on: "http://localhost:3000"
 
       - name: build
         run: |
-          sed -i 's,\(VUE_APP_KEYBOARDS_URL=\).*,\1https://keyboards.develop.qmk.fm,' .env.production
-          echo VUE_APP_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
+          sed -i 's,\(VITE_APP_KEYBOARDS_URL=\).*,\1https://keyboards.develop.qmk.fm,' .env.production
+          echo VITE_APP_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
```
noroadsleft, line 51: Looks to me like this should be

```suggestion
          sed -i 's,\(VITE_KEYBOARDS_URL=\).*,\1https://keyboards.develop.qmk.fm,' .env.production
          echo VITE_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
```

zekth: https://vitejs.dev/guide/env-and-mode.html#modes no

.env
```diff
@@ -1,2 +1,2 @@
-VUE_APP_API_URL=https://api.qmk.fm
-VUE_APP_KEYBOARDS_URL=https://keyboards.qmk.fm
+VITE_API_URL=https://api.qmk.fm
```
zekth, line 1:
```suggestion
VITE_APP_API_URL=https://api.qmk.fm
```

.env
```diff
@@ -1,2 +1,2 @@
...
+VITE_KEYBOARDS_URL=https://keyboards.qmk.fm
```
zekth, line 2:
```suggestion
VITE_APP_KEYBOARDS_URL=https://keyboards.qmk.fm
```

.github/workflows/build.yml
```diff
@@ -33,11 +35,12 @@ jobs:
...
         with:
-          start: yarn serve
-          wait-on: 'http://localhost:8080'
+          cache-key: node-on-${{ runner.os }}-hash-${{ hashFiles('yarn.lock') }}
+          start: yarn dev
+          wait-on: 'http://localhost:3000'
       - name: build
         run: |
-          echo VUE_APP_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
+          echo VITE_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
```
zekth, line 43:
```suggestion
          echo VITE_APP_TRAVIS_COMMIT=$GITHUB_SHA >> .env.production
```

Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR /qmk_configurator/
 
 # Build configurator
 RUN yarn install
-ENV VUE_APP_API_URL=/api
+ENV VITE_API_URL=/api
```
zekth, line 14:
```suggestion
ENV VITE_APP_API_URL=/api
```

index.html
```diff
@@ -5,19 +5,19 @@
...
           margin-left: -100vh;
         }
     </style>
  </head>
-  </head>
   <body>
     <noscript>
       <strong>We're sorry but configurator doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
     </noscript>
+    <script type="module" src="/src/main.js"></script>
```
zekth, line 20: relying directly on /src? Isn't it bundled? i'm not familiar with vite.

zekth (changes requested):
Minor comments

.github/workflows/deploy_develop.yml
noroadsleft: I guess this whole thing is over my head, because nothing on that page explains to me why not. But I trust your opinion on this more than my own.

.env
yanfali: I think documentation says it's prefix is https://vitejs.dev/config/#envprefix am I reading it wrong?

index.html
yanfali: https://vitejs.dev/guide/#index-html-and-project-root

```
Vite treats index.html as source code and part of the module graph. It resolves <script type="module" src="..."> that references your JavaScript source code. Even inline <script type="module"> and CSS referenced via <link href> also enjoy Vite-specific features. In addition, URLs inside index.html are automatically rebased so there's no need for special %PUBLIC_URL% placeholders.
```

yanfali: This is required for ESM module stuff to work in browser for dev mode.

The bundling is done by rollup afaict. In the final dist it writes this:
```
   <script type="module" crossorigin src="/assets/index.dd94f9d5.js"></script>
   <link rel="modulepreload" href="/assets/vendor.61906764.js">
   <link rel="stylesheet" href="/assets/index.1270ddda.css">
```

yanfali:
Thanks guys for being so thorough. I makes me happy you are paying attention to this stuff.

.env
zekth: you're right i was confused by this one: https://vitejs.dev/guide/env-and-mode.html#modes

index.html
zekth: ok fair :)

zekth (approved):

.env
yanfali: I know the doc is really confusing. I think the namespace would be APP_* which in itself isn't a terrible idea.

PR #1068, opened 2022-02-03
changed close to a button

NAXLAB:
Changed the close import button from an anchor to a button


yanfali:
I think we need to fix the formatting. You can fix it in vsc with Ctrl + f and pick prettier or run yarn prettier:fix to do it automatically

PR #1097, opened 2022-03-07
Enable Italian i18n translation in UI

yanfali:
- add italian to app state so we can select it in drop down

Enables the Italian language translation in the drop down menu

## Description

Does this look alright @federikovi ?


federikovi:
@yanfali Looks great to me!

yanfali:
> Are there other things I could translate into Italian?

Have your looked at docs.qmk.fm?

federikovi:
That's the next step

PR #1112, opened 2022-04-19
Updated default keymap for num lock and scroll lock

DeeDesired:
## Description

Num lock and scroll lock showed up as the any key with the key code, so the new version has the proper keycode for num lock and scroll lock.


noroadsleft (approved):

noroadsleft:
Thanks!

noroadsleft:
@DeeDesired,

I'm investigating your Configurator implementation, and there's a situation that isn't clear to me.

The `dee` and `iso` keymaps submitted to [`qmk_firmware`](https://github.com/qmk/qmk_firmware/tree/2714c70bd774489f039b7c60a2f8634a64ff40c3/keyboards/crimsonkeyboards/resume1800/keymaps/) both use the `LAYOUT_resume1800_iso_all` macro, but assign the keys in a different order to each other, which means one of the keymaps will not work as intended when flashed to the board.

Here's how those keymaps look when implemented in Configurator:

`crimsonkeyboards/resume1800:dee`:
crimsonkeyboards_resume1800_dee

`crimsonkeyboards/resume1800:iso`:
crimsonkeyboards_resume1800_iso

Does the `iso` work as intended when flashed to the board? I'm thinking not; but either way, fixing this will require changes on the `qmk_firmware` repo.

Thanks.

PR #1138, opened 2022-06-04
Color Contrast Updates for Nested Keycodes: SA Edition

noroadsleft:
## Description

More of the same. All of the changes here except for SA Bliss are actually contrast reductions to be less jarring.

SA Bliss on Ergodox EZ. Left half current `master`, right half with this PR:

sa-bliss

### DIFF
--- a/src/scss/colorways/_sp.scss
+++ b/src/scss/colorways/_sp.scss
@@ -287,20 +287,24 @@
   background: $color-sp-abs-WV;
   color: $color-sp-abs-BBI;
   input,
-  .key-contents {
-    background: lighten($color-sp-abs-WV, 10%);
+  .key-contents,
+  .key-contents:empty,
+  .key-contents::before {
+    background: lighten($color-sp-abs-WV, 4%);
     color: $color-sp-abs-BBI;
-    border-color: mix($color-sp-abs-WV, $color-sp-abs-BBI);
+    border-color: mix(lighten($color-sp-abs-WV, 4%), $color-sp-abs-BBI);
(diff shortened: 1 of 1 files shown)


yanfali:
Quick question. Does it make a difference if your monitor is uncalibrated? Not questioning your choices, just interested in your opinion. Color and brightness are very subjective and equipment dependent. Otherwise grateful you are doing this work?

yanfali (approved):

noroadsleft:
> Does it make a difference if your monitor is uncalibrated?

Based on my research, not a whole lot. Most modern monitors have decent calibration out of the box, and as you've said, this is a highly subjective area.

I actually turned my monitor brightness down quite a bit when I set it up, and have left it since that initial change. My personal vision is actually fairly sensitive to light and changes in light - so much so that I actually find most screens to be painfully bright at the typical viewing distance of a PC monitor or laptop screen. I actually run f.lux on my PC, with a daylight setting of 4200K (Fluorescent), sunset at 2700K (Incandescent), and bedtime at 1900K (Candle), which I would consider quite low, but this is anecdotal.

The contrast changes have been borne out of my coming to the opinion that my original contrast changes were too stark when comparing a [base key to its inset key](https://user-images.githubusercontent.com/18669334/172024062-7659f7c1-c0ce-4df3-b5c4-5b49885641fc.png) (shown on SA Danger Zone from current `master`). I've come to the opinion that such a wide difference actually makes the color scheme, for lack of a better descriptor, less visually appealing.

PR #1141, opened 2022-06-16
Bubble75.json

Swookian:
Added Bubble75 default configuration. Json provided by Velocifire


noroadsleft (changes requested):
This is not the correct file. This is a VIA JSON file, which is consumed by the VIA client through its API so the client knows the physical layouts that are supported, and the wiring of those layouts.

QMK Configurator's JSON format is completely different. Please follow the documentation at [Adding Default Keymaps to QMK Configurator](https://docs.qmk.fm/#/configurator_default_keymaps).

Please note that QMK Configurator only supports keyboards which have had their support merged into QMK Firmware's mainline at [https://github.com/qmk/qmk_firmware/tree/master/keyboards](https://github.com/qmk/qmk_firmware/tree/master/keyboards).

Issue #866, opened 2020-12-29
Feature Request: Add European layouts to the onscreen keyboard


Abraxas-Knister:
I want to setup a keyboardio/atreus with a neo2 inspired layout, and since I'm German it should be able to actually type the included German ÄäÖöÜüß (capital ß is included in actual neo2 but not required for me).

If I understood the configuration process correctly, I think I need to (a) tell the OS that my external keyboard is intended to be mapped as a qwertz keyboard and (b) configure my layers to use key (combinations) that would create the correct result on a German qwertz keyboard.

For that to be a pleasant process it would be nice to have the option to have the ANSI or ISO onscreen keyboards in the configurator appear with labels matching the target keymapping in the OS.

Examples:

- to type '>' on the ISO qwertz layout you press the key right of the left shift (labeled '\|' in the configurator) together with shift.
- to type ']' on the qwertz layout you press the number row 9 together with AltGr (which hopefully corresponds to right alt), but 9 is labeled '9(' in the configurator
- to get '\' you press AltGr together with the key right of the 0 in the number row, which is '-_' in the configurator.

Also you're probably easily tempted to just drag and drop the '{' from the bottom row in the configurator, but that wouldn't work as expected: it stands for shift plus '[', where with '[' the key to the right of 'p' is meant, which is 'ü' on a qwertz layout. So from '{' you get 'Ü'.

It would therefore be nice to have fast access to common punctuation or symbols for the european qwerty layouts too, or at least have access to them removed.

yanfali:
As has probably been explained to you on discord, all keyboards are QWERTY/ANSI by default on the wire. All layouts are a function of the operating system and vary from OS to OS when you set the locale. It would be quite a lot of work to make this work in the UI. Here's the mappings we provide for the compiler per locale

https://github.com/qmk/qmk_firmware/tree/master/quantum/keymap_extras

To give you a taste of how much work this would actually be.

Abraxas-Knister:
I didn't visit the discord, should I have?

Firstly a question to make sure I understood the mapping process correctly: will my atreus generate an 'ö' on the screen when I set the key below the home position of then left ring finger to generate an '; :',  and will it genereate an '{' if I set the relevant key to right alt plus number row 7 (given I have qwertz in the OS)?

So what would need to be done is

- give the user a menu for choosing the layout
- map the layout in that menu to the correct header
- load the header into the firmware generator
- display the correct layout (ie make little icons for the keys)
[...]

Issue #846, opened 2020-11-18
[Feature Request] Support for other software keyboard layouts


Ingvix:
I think it'd be nice to be able so see the keys represented as they are in the non-ANSI keyboard layout I use on my devices.

Someone hinted that Freedesktop group maintains a pretty comprehensive database of the different keyboard layouts and theoretically it could be queried to provide the functionality to select according to which layout should the keycodes be visually represented in QMK Configurator.

yanfali:
This is a surprisingly complicated task. Basically all keyboards, at the firmware level are actually ANSI over HID USB. The OS is actually responsible for translating those key positions into the locale for you, so we could take the translation header files in theory and provide a mapping, but it's really a function of your operating system, not the keyboard firmware as what the meaning of each of those keycodes is.

Ingvix:
I understand and it's not that huge problem to just deal with it. It would just be a nice visual help when using the configurator to see the keys as you've learned to know them rather than checking all the time what was this and that key's equivalent in my local layout. Until I've done it long enough to memorize the standard ANSI layout entirely.

Issue #768, opened 2020-06-29
Unable to add ISO-specific keys in configurator


IonicOwl:
I'm trying to flesh out an ISO keyboard layout on the configurator, but despite using the keys in the ISO/JIS tab, when I drag them up to the keymap, they show the ANSI key instead. 

In the example below, dragging the KC_2 key up to its position on the top row shows the "2/@" key, instead of the "2/"" key that I'd expect.

![image](https://user-images.githubusercontent.com/36578130/85980714-d4a7d780-b9da-11ea-82a0-1587ed71f3e8.png)

This also occurs with the KC_3 (should be "3/£), KC_GRV and KC_QUOT keys.

Also, since the AltGr key isn't under the ISO tab, am I right in assuming that I should be using the RALT(kc) key under the Quantum tab for accessing graph/accent keys?

yanfali:
@IonicOwl this is a good catch, it's more of a visual oversight than an actual bug. Your operating system is responsible for mapping the ANSI based HID codes onto the locale. I think to fix this it mostly requires fixing up the keycodes for the ISO keymap programmatically to display the correct codes. No actual keycodes need to change.

yanfali:
Can you send a list of the keys that need updating?

noroadsleft:
> Also, since the AltGr key isn't under the ISO tab, am I right in assuming that I should be using the RALT(kc) key under the Quantum tab for accessing graph/accent keys?

Actually, I noticed that the AltGr key was missing from the ISO tab a while back, and I think I may have a branch that adds it.

Gimme some time and I'll dive in to that. The rest of this is as yanfali said – a visual oversight, rather than a bug.

IonicOwl:
> Can you send a list of the keys that need updating? or are the ones in the bug exhaustive?

Those are the only ones that need adding for my layout.  I'm only building a 60% and I'm pretty sure everything on the navigation and number clusters are identical between ISO and ANSI layouts anyway.

noroadsleft:
#770 will add the Alt Gr key to the ISO tab.

Issue #229, opened 2018-12-25
Feature: Non--US Keyboard support


drashna:
Specifically, something that we could toggle to enable non-US layouts, like we do with the `keymap_extras` folder in QMK Firmware. 

That way, for non-US English users, it's more obvious what should go where.

yanfali:
Sorry, can you give me some examples of how that would work?

heieisch:
I looked into the  >iso-jis.js   File and modified a version for German Keybords.
For testing and strange situations it wold be super cool if we coud upload a .json File in the same stucture as the .js Files are now 
In the Future ther coud be an dropdown menu like the keybord seleklektion for custom buttons

yanfali:
@heieisch hi, so this feature request isn't quite what you are requesting. Drashna is looking for API support to include non-us layouts support in QMK header files, which is something we don't support through the API yet. If you want a german ISO layout in the tabs, we could look into that. Attach your copy to this issue. I think a drop down would probably work, where we let you switch between JIS and DE. The other thing we could do is show you de by default if we detect you are from Germany.

One thing we could use help with is a German translation of the UI if you are interested.

yanfali:
@heieisch I would recommend you check out a copy of the repo and install a development environment. then you can do development directly.

heieisch:
I think a drop-down menu would be better due to the fact that many German Programmers code on us layout.

I`m not really good at coding... but I will try my best
is there a guide to set up the development environment for the qmk Configurator?

If there are language files(configuration Files where there is the English version on one side and the translation next to it or something similar), I am going to Provide them. If I have to write code to localize I don´t think it is worth the effort because most Germans know enough English.

heieisch:
[din.zip](https://github.com/qmk/qmk_configurator/files/3214204/din.zip)

heieisch:
I checked this configuration for the last hour and am fairly confident that it might work.
[din.zip](https://github.com/qmk/qmk_configurator/files/3214273/din.zip)

yanfali:
@heieisch Hi! Vielen Dank. So, take a look at this commit https://github.com/qmk/qmk_configurator/commit/f033283a1deaabf77446d090b832dba039acd9b8 This was the translation done very generously by @gab-a  and is a good overview of what is required. It's mostly taking the en.js JSON code and translating the labels into German. If you feel up to the task, we would love a contribution.

Anything you don't translate will fall back to English. So for example you don't have to do the potato facts.

[...]

PR #1161, opened 2022-08-08
Add support for intl key legends based on the host keyboard layout

precondition:
## Description
[demo-intl-configurator.webm](https://user-images.githubusercontent.com/57645186/183497044-4ac487c7-df29-4768-bb6c-cd585779d08c.webm)

(I capped the framerate at 20fps to reduce the file size, that's why it looks choppy)

I replaced the ISO toggle switch in the settings by a more general dropdown that lets the users pick the correct logical/functional keyboard layout of their OS. The only effect of this setting is to change how each keycode is displayed; nothing about the constructed `keymap.json` is altered.

To get started, I added the following host keyboard layouts:

- US
- UK
- German
- Russian

The lookup tables used for those host keyboard layouts can be found in `src/i18n/keymap_extras` and were generated with the help of the `src/i18n/keymap_extras/convert_keymap_extras_header.js` script. The script would allow me to easily add support for every keymap_extra input language that qmk_firmware supports, but for the sake of keeping this PR more concise, I limited myself to four OS keyboard layouts.

The bulk of the relatively impressive amount of added lines in this PR comes from those generated lookup tables, and to a lesser extend from the optional converter script used to generate them. The actual core changes to qmk_configurator are small.

## Future Work
Can be addressed in follow-up PRs

- [x] Support more OS layouts (See #1165)
- [ ] Add the corresponding `"host_language"` in the `keymap.json`
- [ ] Guess the user's OS keyboard layout based on their locale
- [ ] Possibly hide keys for capital letters in the "Shifted Symbols" section
- [x] Include comments and ASCII art in the generated output of `convert_keymap_extras.py`
- [ ] Document the process of adding support for a new keymap_extras/OS keyboard layout
- [x] Sort the OS kb layout drop-down menu by localised/translated labels (Done in https://github.com/qmk/qmk_configurator/pull/1161/commits/cc349e1f22b4f84cb66924ce5c6f7d8e6b90ea99)
- [ ] Remove all dead code related to the defunct ISO toggle
- [ ] Support localisation for the magic swap keycodes in the `convert_keymap_extras_header.js` script

## Known Bugs

- [x] The text size for legends is computed based on the length of the key.name of the OS keyboard layout initially loaded and does not get updated when switching OS layouts until the user initiates a refresh. Concretely, this means that if the last saved OS layout was keymap_us, the name property of `KC_SCLN` would have been ":\n;". If the user switches to keymap_german, the locale keysym of `KC_SCLN` ("ö") will be shown in small text despite being a single character. ~~(Solved in b84a67513c22c4d7f93a948dddf1edd52330978e)~~ Indirectly solved by #1166
- [x] The test page (config.qmk.fm/#/test) doesn't dynamically update the key legends when switching OS layouts in the menu. You need to switch and reload the page for the new OS keyboard layout to apply. ~~(Solved in 1bad93e)~~ Indirectly solved by #1166
- [x] Host Keyboard Layout dropdown is empty if an old configuratorSettings object missing the new `osKeyboardLayout` property is in localstorage. (Solved in 82cb0332a255a51204da27eddf5ea3b6279e479d)

## Motivation

Hello, I'm using an AZERTY layout and im wondering how can I set my < and > on the keyboard, mapping kc_lt & kc_gt doesn't work and puts . and / insead.
Any ideas on how can I solve this ? thanks

![Heya folks! Managed to flash my keyboard, and everything seems to work just fine. QMK seems powerful and great!
I've been using https://config.qmk.fm/ to configure the layout, but what layout does that site assume my OS uses?
For instance: Assuming I flash with the default layout and my OS's layout is set to "US-QWERTY", pressing the key z key on my keyboard would produce a z in the text editor. However, if I now change my OS's layout to "DE-QWERTZ" (which swaps the z and y keys), if I press the same key, it would produce a y instead.
That means that, when the user does not set the correct layout, the layout they just flashed won't entirely function as config.qmk.fm intended or displayed it to. So what is that 'correct layout'?
Related to that: Assuming I would like to build my keyboards in respect to "DE-QWERTZ" (for instance, if one visits public libraries in Germany a lot and would like to take their keyboard there), can I still use config.qmk.fm? That is, is there some setting or application that can convert from the 'correct layout' to any other layout?](https://i.imgur.com/FjyFXJj.png)

my keymap is great and everything but i was just wondering if i could export it from what it is rn because some keys are out of place on it like the apostrophe and the { key are switched and the colon and semicolon button is ñ 
besides that everything else is in the right place
i just dont know why it would be those keys instead of theyre intended one

### Linked Issues

Feature Request: Add European layouts to the onscreen keyboard #866

[Feature Request] Support for other software keyboard layouts #846 

Unable to add ISO-specific keys in configurator #768

Feature: Non--US Keyboard support #229


src/components/SettingsPanel.vue
```diff
@@ -91,6 +91,28 @@
...
+      <div>
+        <select
+          id="setting-panel-host-keyboard-layout"
+          v-model="hostKeyboardLayout"
+        >
+          <option
+            v-for="osLayout in hostKeyboardLayouts"
+            :key="osLayout"
+            :value="osLayout"
+          >
```
precondition, line 111: It would have been more readable in my opinion to keep it the options of a tag in a single line but I can't commit without calling the linter ¯\_ (ツ)_/¯

src/store/modules/keycodes/index.js
```diff
@@ -27,17 +28,55 @@ const state = {
...
-  return keycodes;
+  if (!Object.keys(keymapExtras).includes(getOSKeyboardLayout())) {
+    console.log(
+      `${getOSKeyboardLayout()} is not a valid OS keyboard layout value!`
+    );
+    return keycodes;
+  }
+  return keycodes.map((keycodeObject) =>
+    toLocaleKeycode(
+      keymapExtras[getOSKeyboardLayout()].keycodeLUT,
```
yanfali, line 76: hoist this line out of map. You're calling a function and then de-referencing the same data every invocation of map. That shouldn't change ever while you are inside the loop.

yanfali:
I'm still digesting the PR, so I'm not ignoring you I promise. I think it looks pretty good, I just want to think about it.

I'm not thrilled to add python to the repo as it's even less portable than node in some respects.

noroadsleft (commented):
Likewise, I'm not ignoring you or this PR either; unfortunately you've caught me just a few days ahead of being out of town for six days. I probably won't be able to properly review this until next week.

yanfali:
> Sometimes, switching from an OS layout where a key has a shift pair to an OS layout where that same key is just a letter displays the letter in small script because it still thinks that the length of key.name is greater than 1. A refresh fixes it.
This may be related to how the key name is being updated. The UI won't render if the reference doesn't change, depending on how it's rendering. I haven't looked at this code in a long time, It might be as simple as re-assigning the object to a new one and copying all the fields.

precondition:
> I'm not thrilled to add python to the repo as it's even less portable than node in some respects.

I can translate the script into node.js if needs be. It is mostly text and hashmap manipulation so it is not hard to port.

src/store/modules/keycodes/index.js
precondition: Solved in https://github.com/qmk/qmk_configurator/pull/1161/commits/bbd6dc9c3b4361b52932eceb776990340280d1ce

src/i18n/keymap_extras/index.js
```diff
@@ -0,0 +1,60 @@
...
+    sendstring: '',
+    isANSI: true,
+    locales: ['en-US', 'en'],
+    keycodeLUT: us
+  },
+  keymap_uk: {
+    prefix: 'UK',
+    sendstring: 'uk',
+    isANSI: false,
+    locales: ['en-UK', 'en-GB', 'en'],
```
precondition, line 39: I'm not sure if 'en-UK' is even a possible locale value for web navigator.

precondition:
If we ignore the deprecated keymap_nordic.h and the steno/plover header files that are out-of-scope, the converter script correctly generates output for every<sup>†</sup> keymap_*.h header in qmk_firmware/quantum/keymap_extras!

<sup>†</sup>: keymap_bepo.h and keymap_lithuanian_qwerty.h have bugs but I already opened PRs to fix them upstream (https://github.com/qmk/qmk_firmware/pull/17999 & https://github.com/qmk/qmk_firmware/pull/18000)

src/i18n/keymap_extras/index.js
yanfali: @zvecr any idea?

yanfali:
> If we ignore the deprecated keymap_nordic.h and the steno/plover header files that are out-of-scope, the converter script generates (syntaxically) correct output for every† keymap_*.h header in qmk_firmware/quantum/keymap_extras!
> 
> †: Assuming that one merges the PRs I opened to fix the bugs in keymap_bepo.h ([qmk/qmk_firmware#17999](https://github.com/qmk/qmk_firmware/pull/17999)) and keymap_lithuanian_qwerty.h ([qmk/qmk_firmware#18000](https://github.com/qmk/qmk_firmware/pull/18000))

Thanks for doing this. I am traveling at the moment, but I will take a look this weekend. I need to test it myself.

src/i18n/keymap_extras/convert_keymap_extras_header.js
```diff
@@ -0,0 +1,300 @@
...
+      //
+      // For an example of a keymap_extras header containing those characters, check out keymap_bepo.h
+      const dottedCircle = '\u25cc';
+      readableKeysym = dottedCircle + keysym;
+    }
+    // aka the hover tip
+    const title =
+      intlAlias + (comment === undefined ? '' : ' ' + comment.trim());
+    kcInfo.set(translateToUS(macroExpansion, intl2us), {
+      intlAlias: intlAlias,
```
yanfali, line 160: you can just use `intlAlias` and it will include it as part of object creating syntax. same goes for comment, and title.

src/i18n/keymap_extras/index.js
zvecr: Only ever encountered `en-gb`, though take that with a pinch of salt.

src/i18n/keymap_extras/convert_keymap_extras_header.js
precondition: Solved in https://github.com/qmk/qmk_configurator/pull/1161/commits/16ee0e1116b6aed69f0365e5b47234cb5570892e

precondition:
While working on localising `QK_GESC`, I realised that its wording can be misleading. "Esc normally, but ~ when Shift or GUI is pressed". First of all, the Shift or GUI modifiers don't necessarily need to be pressed while pressing `QK_GESC`, it would work just fine even if you tapped a one-shot modifier. The important point is that one of those modifiers needs to be *active*. Secondly, when GUI is active, `QK_GESC` produces `\`` (at least on US), not `~`.

My suggestion:
```diff
diff --git a/src/store/modules/keycodes/quantum.js b/src/store/modules/keycodes/quantum.js
index 734e9a1..90c5921 100644
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -392,9 +392,9 @@ export default [
   { label: 'Special action keys', width: 'label' },
 
   {
-    name: 'Esc / ~',
+    name: '` / ~\nEsc',
     code: 'QK_GESC',
-    title: 'Esc normally, but ~ when Shift or GUI is pressed'
+    title: 'Esc normally, but ` when GUI is active or ~ when Shift is active'
   },
   {
     name: 'LS / (',
 
```
image

Displaying the legend on 2 lines would also be more coherent with the rest of the shift pairs. If you think about it, the shift pair of `QK_GESC` is base=Esc and Shift=~ after all.

yanfali:
> While working on localising `QK_GESC`, I realised that its wording can be misleading. "Esc normally, but ~ when Shift or GUI is pressed". First of all, the Shift or GUI modifiers don't necessarily need to be pressed while pressing `QK_GESC`, it would work just fine even if you tapped a one-shot modifier. The important point is that one of those modifiers needs to be _active_. Secondly, when GUI is active, `QK_GESC` produces `` ` `` (at least on US), not `~`.
> 
> My suggestion:
> 
> ```diff
> diff --git a/src/store/modules/keycodes/quantum.js b/src/store/modules/keycodes/quantum.js
> index 734e9a1..90c5921 100644
> --- a/src/store/modules/keycodes/quantum.js
> +++ b/src/store/modules/keycodes/quantum.js
> @@ -392,9 +392,9 @@ export default [
>    { label: 'Special action keys', width: 'label' },
>  
>    {
> -    name: 'Esc / ~',
> +    name: '` / ~\nEsc',
>      code: 'QK_GESC',
> -    title: 'Esc normally, but ~ when Shift or GUI is pressed'
> +    title: 'Esc normally, but ` when GUI is active or ~ when Shift is active'
>    },
>    {
>      name: 'LS / (',
>  
> ```
> 
> image
> 
> Displaying the legend on 2 lines would also be more coherent with the rest of the shift pairs. If you think about it, the shift pair of `QK_GESC` is base=`Esc` and Shift=`~` after all. Although I'm not entirely convinced by the ` / ~` part.
> 
> See commit [5b36bf9](https://github.com/qmk/qmk_configurator/commit/5b36bf9d6424aa8926b9f38c193078771b03ff64)

I have no opinion on this, as it doesn't work on macOS at all due to a different binding policy. @qmk/collaborators may have other opinions.

precondition:
I just realised that since I mostly rely on the old pre-existing ISO toggle mechanism to update the key legends text, the test page (config.qmk.fm/#/test) doesn't dynamically update the key legends when switching OS layouts in the menu (since the old ISO toggle has no effect on that page). You need to switch and reload the page for the new OS keyboard layout to apply.

I'm in the middle of a move right now and I will not have Wi-Fi in the next few days so I don't know when I'll be able to look into the issue.

yanfali:
> I just realised that since I mostly rely on the old pre-existing ISO toggle mechanism to update the key legends text, the test page (config.qmk.fm/#/test) doesn't dynamically update the key legends when switching OS layouts in the menu (since the old ISO toggle has no effect on that page). You need to switch and reload the page for the new OS keyboard layout to apply.
> 
> I have a feeling that fixing the first known bug will also patch this. However, I'm in the middle of a move right now and I will not have Wi-Fi in the next few days so I don't know when I'll be able to look into the issue.

This never needed to work iirc as you could 

> I just realised that since I mostly rely on the old pre-existing ISO toggle mechanism to update the key legends text, the test page (config.qmk.fm/#/test) doesn't dynamically update the key legends when switching OS layouts in the menu (since the old ISO toggle has no effect on that page). You need to switch and reload the page for the new OS keyboard layout to apply.
> 
> I have a feeling that fixing the first known bug will also patch this. However, I'm in the middle of a move right now and I will not have Wi-Fi in the next few days so I don't know when I'll be able to look into the issue.

As there are explicit ISO and ANSI buttons it was never really a priority to make this toggle either. It wouldn't be too hard to fix this though.

yanfali:
super busy at work. I'll try and make time for this on thursday

src/i18n/keymap_extras/index.js
noroadsleft: Any usage of `en-UK` would be wrong; the officially-designated IETF language tag for British English is `en-GB`.

precondition: Ok, removed in ac3f3c216ee486d8ef7193fe29

yanfali:
Thank you for all the hardwork. I will be reviewing this properly today, evening PST.

precondition:
As I've guessed, the two known bugs share a common root cause. The root cause is that the values of `this.meta` (and by extension, `this.meta.name`) are set according to the OS keyboard layout selected during the initial loading of the web page, and does not get updated when the user switches OS layouts in the settings panel. Consequently, any components like BaseKey.vue or TesterKey.vue that rely on the value of `this.meta.name` to carry out their tasks do not react appropriately to the new OS layout settings (until a refresh of the page is made).

My solution, as presented in commits https://github.com/qmk/qmk_configurator/pull/1161/commits/b84a67513c22c4d7f93a948dddf1edd52330978e and https://github.com/qmk/qmk_configurator/pull/1161/commits/1bad93e13014dd220f487ba042edd4dfeed59347, is to use `const { name } = lookupKeycode(this.meta.code);` instead of `this.meta.name`. It works but I'm not a big fan of paying the *O*(n) price of the `lookupKeycode` call (due to `state.keycodes.find()`). Ideally, `this.meta.name` should get updated when the user switches OS keyboard layouts so that we can stay in *O*(1).

yanfali:
There's an issue for the very first time it loads, the localstorage stuff is out of sync until the first reload.

src/store/modules/app/mutations.js
```diff
@@ -181,6 +181,9 @@ const mutations = {
   setDarkmode(state, value) {
     state.configuratorSettings.darkmodeEnabled = value;
   },
+  setOSKeyboardLayout(state, value) {
+    state.configuratorSettings.osKeyboardLayout = value;
```
yanfali, line 185: As you added a new configuratorSettings key osKeyboardLayout, but state.js doesn't know about it. You should add this to intialConfig and give it a default otherwise the UI is a bit wiggy and we need to rev the localStorage version.

yanfali: The default should be US english as that is the existing behavior.

yanfali:
when switching to russian the ISO big enter doesn't kick in. Is that the expected behavior?

src/store/modules/app/mutations.js
precondition: That's strange because I did define the default osKeyboardLayout value in the initialConfig: https://github.com/qmk/qmk_configurator/blob/0a235d32896d9ba1e869da22c0a217c1d439fa2d/src/store/modules/app/state.js#L19

precondition:
> when switching to russian the ISO big enter doesn't kick in. Is that the expected behavior?

Yes, the Russian keymap is set to be ANSI because the OS layout produces the same keysym for `KC_BSLS`, `KC_NUHS` and `KC_NUBS`. Additionally, the `keymap_russian.h` file included in qmk_firmware clearly shows an ANSI ASCII art:
```c
/*
 * ┌───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───┬───────┐
 * │ Ё │ 1 │ 2 │ 3 │ 4 │ 5 │ 6 │ 7 │ 8 │ 9 │ 0 │ - │ = │       │
 * ├───┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─────┤
 * │     │ Й │ Ц │ У │ К │ Е │ Н │ Г │ Ш │ Щ │ З │ Х │ Ъ │  \  │
 * ├─────┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴┬──┴─────┤
 * │      │ Ф │ Ы │ В │ А │ П │ Р │ О │ Л │ Д │ Ж │ Э │        │
 * ├──────┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴─┬─┴────────┤
 * │        │ Я │ Ч │ С │ М │ И │ Т │ Ь │ Б │ Ю │ . │          │
 * ├────┬───┴┬──┴─┬─┴───┴───┴───┴───┴───┴──┬┴───┼───┴┬────┬────┤
 * │    │    │    │                        │    │    │    │    │
 * └────┴────┴────┴────────────────────────┴────┴────┴────┴────┘
 */
```

yanfali:
OK, back to this KC_SCLN bug. Can you give me screenshots as to what the behavior is without the patch? I tried reproducing it but since I'm not familiar with ISO layout I can't tell if you're talking about the keycodes keyboard on the main keymap display as to how the bug happens. Sorry for being slow about this, but I would like to see what's happening so I can debug it and maybe give you a less janky fix. I'm guessing you're talking about the keymap.

So some useful things, what keyboard are you testing with, so I can see the behavior and steps to reproduce. Thanks

yanfali:
Ah ok. I think I reproduced it. I see the smaller css attached to the key rendering now.

precondition:
> OK, back to this KC_SCLN bug. Can you give me screenshots as to what the behavior is without the patch? I tried reproducing it but since I'm not familiar with ISO layout I can't tell if you're talking about the keycodes keyboard on the main keymap display as to how the bug happens. Sorry for being slow about this, but I would like to see what's happening so I can debug it and maybe give you a less janky fix. I'm guessing you're talking about the keymap.
> 
> So some useful things, what keyboard are you testing with, so I can see the behavior and steps to reproduce. Thanks

Since the root cause is the same, you can also check http://localhost:3000/#/test (or whatever localhost port your local dev env is hosted in). It is a lot easier to notice; without the patch, switching the OS keyboard layout in the settings panel has no effect on the key legends and requires a page refresh (since it's reading `this.meta.name`)

src/store/modules/app/mutations.js
noroadsleft: I finally got around to testing this PR and I had the same issue of `osKeyboardLayout` being unset initially.

noroadsleft: And of course right after I say that, I have a theory to test, which I tested and proved.

The problem here is that `initialConfig` is only used if there are no settings saved in Local Storage at all. If you open your development console in the browser, delete `configuratorSettings` from Local Storage, then refresh the page, it works as expected.

We probably need a check to see if there's an existing setting for `osKeyboardLayout` in `configuratorSettings`, in addition to checking if there are any settings saved at all.

noroadsleft:
Needs a rebase.

src/store/modules/app/mutations.js
precondition: > And of course right after I say that, I have a theory to test, which I tested and proved.
> 
> The problem here is that `initialConfig` is only used if there are no settings saved in Local Storage at all. If you open your development console in the browser, delete `configuratorSettings` from Local Storage, then refresh the page, it works as expected.
> 
> We probably need a check to see if there's an existing setting for `osKeyboardLayout` in `configuratorSettings`, in addition to checking if there are any settings saved at all.

I managed to reproduce the problem and issue a fix in commit 566678a

src/components/VisualTesterKeymap.vue
```diff
@@ -1,6 +1,17 @@
 <template>
   <div class="tester">
-    <div class="layout-selector-radios"></div>
```
precondition, line 3: I know that I'm undoing your work, @yanfali, but in the absence of a ISO toggle in the settings panel, users should still be able to switch keyboard geometries without having to change the selected OS keyboard layout.

noroadsleft:
Full disclosure: I just hit a weird UI bug, running on 566678a0509e762358c3216adddc3d87630e14c0.

It seems to be contingent on:

- `configuratorSettings` being present 
- `osKeyboardLayout` being *absent*

1. Open Configurator
   - UI opens with `English (US)` defaulted in the Settings panel, but `osKeyboardLayout` not set.
2. Open the Settings panel
3. Change from `English (US)` to any other option, but *leave the panel open*
   - Keymap updates; `osKeyboardLayout` is set

After a couple of seconds, the `Host Keyboard Layout` will revert to its previous setting, but `osKeyboardLayout` will remain unchanged and the keymap will remain in whichever layout was selected in Step 3.

Refreshing the page seems to fix this (because `osKeyboardLayout` is no longer absent on first load).

precondition:
> Full disclosure: I just hit a weird UI bug, running on [566678a](https://github.com/qmk/qmk_configurator/commit/566678a0509e762358c3216adddc3d87630e14c0).
> 
> It seems to be contingent on:
> 
>     * `configuratorSettings` being present
> 
>     * `osKeyboardLayout` being _absent_
> 
> 
>     1. Open Configurator
>        
>        * UI opens with `English (US)` defaulted in the Settings panel, but `osKeyboardLayout` not set.
> 
>     2. Open the Settings panel
> 
>     3. Change from `English (US)` to any other option, but _leave the panel open_
>        
>        * Keymap updates; `osKeyboardLayout` is set
> 
> 
> After a couple of seconds, the `Host Keyboard Layout` drop-down will revert to its previous setting, but `osKeyboardLayout` will remain unchanged and the keymap will remain in whichever layout was selected in Step 3.
> 
> Refreshing the page seems to fix this (because `osKeyboardLayout` is no longer absent on first load).
> 
> Edit: Alternatively, one can close and reopen the Settings panel.

Can you try again with commit ac769986cb2889f0b2e34b0e5f5c9fff70e18a92 ? After this commit, I can't reproduce the issue you describe but it wouldn't be the first time that I wrongly declare that the dropdown is free of bugs so I'm cautious haha

src/components/VisualTesterKeymap.vue
yanfali: Fair enough. I took it out because it seemed weird to not switch if you selected ISO in the settings panel. Now you've added locale it's more nuanced.

src/store/modules/keycodes/index.js
```diff
@@ -76,22 +121,22 @@ const mutations = {
...
-    if (state.active === 'ANSI') {
-      state.active = 'ISO/JIS';
-    }
+    state.keycodes = generateKeycodes(getOSKeyboardLayout(), state.steno);
   },
-  disableIso(state) {
-    state.keycodes = generateKeycodes(isISO(), state.steno);
-    if (state.active === 'ISO/JIS') {
-      state.active = 'ANSI';
+  updateKeycodeNames(state) {
```
yanfali, line 130: I think this needs to go after state.keycodes as it relies on the current value of state.keycodes for the lookup.

yanfali: never mind. I'm misreading the diff.

src/store/modules/keycodes/index.js
```diff
@@ -76,22 +121,22 @@ const mutations = {
...
+    state.keycodes = generateKeycodes(getOSKeyboardLayout(), state.steno);
+    if (isANSI()) {
```
yanfali, line 132: Minor issue, but this state change feels like it should probably be in it's own function. You're changing the state of the keycodes and the active tab.

src/components/SettingsPanel.vue
```diff
@@ -166,6 +171,22 @@ export default {
       async set(value) {
         await this.changeLanguage(value);
       }
+    },
+    osKeyboardLayout: {
+      get() {
+        return this.configuratorSettings.osKeyboardLayout;
+      },
+      async set(value) {
+        await this.changeOSKeyboardLayout(value);
```
yanfali, line 180: Because this is async, I would just wrap this in a try { ... } catch (error) { console.log(error) } because it's async, it could potentially create an unhandled promise exception.

yanfali (commented):
I think this is looking really good. I'll try and re-test it this weekend and I think we can probably get it merged then. Thank you for your hard work and patience.

src/store/modules/keycodes/index.js
precondition: Solved in eaee8fa190c69f6ae9c01cc653b27b3de288211d

src/components/SettingsPanel.vue
precondition: Solved in 414986d

noroadsleft:
> Can you try again with commit ac769986cb2889f0b2e34b0e5f5c9fff70e18a92 ? After this commit, I can't reproduce the issue you describe but it wouldn't be the first time that the drop-down suffers from unforeseen bugs.

> The changes should in theory prevent other similar new-property-missing-from-locally-stored-settings problems in the future.

Issue is fixed. :heavy_check_mark:

src/components/SettingsPanel.vue
```diff
@@ -91,6 +91,28 @@
...
+        <select
+          id="setting-panel-os-keyboard-layout"
+          v-model="osKeyboardLayout"
+        >
+          <option
+            v-for="osLayout in sortedOSKeyboardLayouts"
+            :key="osLayout"
+            :value="osLayout"
+          >
+            {{ $t('settingsPanel.osKeyboardLayout.label.' + osLayout) }}
```
yanfali, line 112: prefer backticks
```javascript
`settingsPanel.osKeyboardLayout.label.${osLayout}`
```

src/store/modules/app/state.js
```diff
@@ -15,21 +15,24 @@ function setDefaultConfiguratorSettings() {
...
   try {
-    let conf = JSON.parse(localStorageLoad(CONSTS.configuratorSettings));
-    if (!conf) {
-      return setDefaultConfiguratorSettings();
-    }
+    // Locally stored configuratorSettings can miss properties that were added in a new update,
+    // so it is important to merge them with the default configurator settings to ensure that all
+    // settings are set.
+    let conf = {
+      ...getDefaultConfiguratorSettings(),
```
yanfali, line 30: nice.

src/store/modules/keycodes/index.js
```diff
@@ -27,17 +31,58 @@ const state = {
...
+  const osKeyboardLayout =
+    store.state.app.configuratorSettings.osKeyboardLayout;
+  if (
+    isUndefined(osKeyboardLayout) ||
+    !Object.keys(keymapExtras).includes(osKeyboardLayout)
+  ) {
+    console.log(
+      `The stored OS keyboard layout value (${osKeyboardLayout}) is not a valid value! Falling back to '${fallbackOSKeyboardLayout}'.`
+    );
+    store.state.app.configuratorSettings.osKeyboardLayout =
```
yanfali, line 45: this isn't really a good idea. we should go through a commit to do this. I'd also use an actual getter on grabbing the state. now you know too much about the internals of store.state.app

src/store/modules/keycodes/index.js
```diff
@@ -27,17 +31,58 @@ const state = {
...
+    // override the name, title, and possibly other fields
+    return { ...keycodeObject, ...keycodeLUT[keycodeObject.code] };
+  } else {
+    return keycodeObject;
+  }
 }
 
-function generateKeycodes(isIso, isSteno) {
+function generateKeycodes(osKeyboardLayout, isSteno) {
+  store.state.app.configuratorSettings.iso = !isANSI();
```
yanfali, line 74: same here. Use an update to configurator settings, don't change it like this directly. Cross store updates like this is buggy and untracked by vuex.

yanfali (changes requested):
A few minor changes need to manipulating the configurator saved state, otherwise LGTM.

yanfali:
@precondition I think we're almost ready to merge this. Thanks again for this. Very cool feature. Just a couple of very small clean ups and we're good to go.

src/components/SettingsPanel.vue
precondition: Solved in fee4d15

src/store/modules/keycodes/index.js
precondition: Solved in 2f5c48a

src/store/modules/keycodes/index.js
precondition: Solved in 8cb10a2

yanfali (approved):
Lgtm. @noroadsleft can you give it one more test. Otherwise I think it's ready to merge.

yanfali:
@noroadsleft bump

noroadsleft (approved):
Looks good to me.

Thanks for this; been wanting this feature for a long time.

precondition:
Should the linked issues be closed now?

PR #1165, opened 2022-08-19
Add more OS keyboard layouts

precondition:
## Description

Generated keycode LUTs for conversion of the configurator front-end from one OS keyboard layout to another with the help of the `convert_keymap_extras_header.js` script added in #1161.


src/store/modules/app/state.js
```diff
@@ -79,10 +79,67 @@ const state = {
     { value: 'zh-CN', label: '简体中文' }
   ],
   osKeyboardLayouts: [
+    'keymap_belgian',
```
yanfali, line 82: ok this is long enough now, please extract it to it's own JS file and import it. :)

precondition: Done in commit ffcfc24. Although I'm not sure what's the best directory to put in or what case mode to use for the filename (`exclusion_list.js`, `longFormKeycodes.js`, and `potato-facts.js` all use a different case mode so I don't know what's the desired style for filenames)

precondition: Ideally, we wouldn't even need to maintain a list of OS keyboard layout identifiers and just tell the configurator to list all `src/i18n/keymap_extras/keymap_*.js` files but I don't know what would be an idiomatic way to achieve that.

yanfali: > Ideally, we wouldn't even need to maintain a list of OS keyboard layout identifiers and just tell the configurator to list all `src/i18n/keymap_extras/keymap_*.js` files (and strip the path stem and .js extension) but I don't know what would be an idiomatic way to achieve that.

We would have to do it at build time and write some node js that run as part of the build and generated the file.

yanfali: > Done in commit [ffcfc24](https://github.com/qmk/qmk_configurator/commit/ffcfc2481f51d27ae6cbf2de2ac91462dfaf39c2). Although I'm not sure what's the best directory to put in or what case mode to use for the filename (`exclusion_list.js`, `longFormKeycodes.js`, and `potato-facts.js` all use a different case mode so I don't know what's the desired style for filenames)

if you want to rename all the files to one of the styles for consistency I'd be open to that. I think I'd prefer the snake casing if we were going to pick a style.

precondition:
Bump

yanfali:
This is kind of big. I know it's a lot of configuration. One concern is that it's loading all this stuff into memory. I wonder if we can lazy load this as it's not needed by 90% of users. I'll look at it over the next few days.

yanfali:
Intending to review this Sunday my time. Busy week.

yanfali (approved):
What a passion project. Very impressive. As far as I can tell it doesn't seem to impact loading times too much as it's mostly text which should compress well.

I happened to find a feature you might like in the vite docs:

https://vitejs.dev/guide/features.html#glob-import

I think we talked about this once a while back.

yanfali:
@noroadsleft give this a once over please. My only usability peccadillo is that the drop down is now huge. But other than that it's a very cool feature.

noroadsleft:
> My only usability peccadillo is that the drop down is now huge.

It's not only huge, it's also wide – nearly wider than the settings panel itself, at least for me:

image

noroadsleft (approved):
Looks good to me.

noroadsleft:
Thanks!

yanfali:
Nah. It's not a feature anyone will use very frequently. It should be fine.

PR #1166, opened 2022-08-21
Explore how to support internationalization of the keymap

yanfali:
This is a bit of a mish mash. Working on issues for preconditions new PR for international support I found many smaller issues with the way I wrote this code, before the arrival of await support. Things were very racey before because it was a pain to write this code using callbacks.

Flattens out some fo the callbacks so they are easier to comprehend.

Also cleans up some issues with visual code so that the warnings are now better and removes a really annoying volar issue with vue2/vue3 complaining about something which isn't valid.

Also fixes the way keymap rendering works, which was the original intent of this work.
So in the original ISO/ANSI support contributed, thanks @itsjxck the choice to re-lookup every keycode in the BaseKey was actually just fine because at the time we only supported 2 layouts, ANSI and ISO. @precondition opened a PR which will enable every layout and this exposed a bug in the original implementation which we did not notice.

Basically many rendering decisions are made based on the value of each keys meta data. Specifically meta.name.
when we switch between different locales, this usually wasn't a problem as the name's didn't change much in length. But, in some locales we go away from multi character rendering towards single character rendering. Because some of the CSS applied is using the length of the name to make class decisions, which would result in some keys with single character names getting rendered as if they were multi-character. 

Example is KC_SCLN, which in german is `ö`, and in ANSI is rendered as `:\n;`.

So this patch changes the logic, so that instead of doing the look up dynamically as we're rendering, when we switch locale just sweep the entire keymap and update all the names with the current name for the current locale.

Why would we do it this way? Well it turns out we also use meta.name is any place we render keys. Like BaseKey, PrintKey and TesterKey.

~~TODO: fix the keyboard tester so this logic also works.~~ Fixed keyboard tester so it reacts to the ISO toggle.


src/components/ControllerTop.vue
```diff
@@ -202,6 +208,11 @@ export default {
       }
     }
   },
+  async mounted() {
+    await this.initializeKeyboards();
```
yanfali, line 212: just make all these tasks synchronous, it doesn't make sense to not wait for all of these things to complete

src/components/VisualTesterKeymap.vue
```diff
@@ -1,24 +1,13 @@
...
-          class="layout-btn-select"
-          v-on:click="layout = _layout"
-          :key="_layout"
-          :class="{ active: _layout === layout }"
-        >
-          {{ _layout }}
-        </button>
-      </slot>
-    </div>
+    <div class="layout-selector-radios"></div>
```
yanfali, line 3: remove buttons but keep layout the same.

src/store/modules/app/actions.js
```diff
@@ -56,47 +56,43 @@ const actions = {
    *  @param {string} keyboard new keyboard we are switching to
    *  @return {object} promise that will be fulfilled once action is complete
    */
-  changeKeyboard({ state, commit, dispatch }, keyboard) {
+  async changeKeyboard({ state, commit, dispatch }, keyboard) {
```
yanfali, line 59: just make this a synchronous function

tests/integration/tester.spec.js
```diff
@@ -22,7 +22,8 @@ describe('Tester feature', function () {
   });
   it('Should change layout', function () {
     // ISO has 105 keys
-    cy.get('.layout-selector-radios > :nth-child(2)').click();
+    cy.get('.bes-controls').click();
```
yanfali, line 25: fix tests for new UI behavior

src/store/modules/app/actions.js
```diff
@@ -162,6 +158,9 @@ const actions = {
     commit('setIso', iso);
     const keyboardLayout = iso ? 'enableIso' : 'disableIso';
     this.commit(`keycodes/${keyboardLayout}`);
+    this.commit('keymap/updateKeycodeNames');
```
yanfali, line 161: new hooks go here for now. They basically force keymap to update all the names, for both the keymap and the tester.

src/store/modules/keymap.js
```diff
@@ -238,6 +237,31 @@ const mutations = {
     }
     state.selectedIndex = index;
   },
+  /**
+   * Use this after switching out the keycodes for a new layout to keep the keymap
+   * in sync. It has it's own copy of the labels, which need to be updated for
+   * correct ISO support.
+   * @param {} state
+   */
+  updateKeycodeNames(state) {
```
yanfali, line 246: So this is the core of the new behavior. When requested, force the keymap to update all the keycode names from the latest lookup keycode table. This is what the UI uses to determine what to display. The keycodes don't change between locales. Only the name/label/displayname which is all derived for name. The only special handling here is for container keys, where it's the contained key that needs an update, not the container.

src/store/modules/keymap.js
```diff
@@ -238,6 +237,31 @@ const mutations = {
...
+    let store = this;
+    // assumes the keycode store has changed due to layout update
+    state.keymap = state.keymap.reduce((layers, layer) => {
```
yanfali, line 249: to ensure reactivity works correctly, just change the reference to a new keymap with all the old values copied into it. This will force the UI to re-render.

src/store/modules/app/actions.js
precondition: I'm confused, why does the code say "async"  but your comment says the opposite?

noroadsleft (approved):
Looks good to me.

src/store/modules/app/actions.js
yanfali: It's the irony of async. It makes async things appear synchronous to the developer. Hence air quotes.

PR #1181, opened 2022-09-04
add tooltips with counts

yanfali:
Just for fun, add a count of number of keyboards returns by API to a tooltip on keyboard.
Also display the number of layouts loaded as a tooltip.

Minor clean ups and upgrades


yanfali:
Screen Shot 2022-09-04 at 00 09 55

src/components/ControllerTop.vue
```diff
@@ -26,7 +29,10 @@
         ></v-select>
       </div>
       <div class="topctrl-layouts">
-        <label id="drop-label-version" class="drop-label"
+        <label
+          id="drop-label-version"
+          v-tooltip.top-end="`${Object.keys(layouts).length} layouts`"
+          class="drop-label"
           >{{ $t('layout.label') }}:</label
```
mateossh, line 36: There's two different closing HTML tags approaches


```html
<elem
  prop1
  prop2
  >.....</elem
>
```

and

```html
<elem
  prop1
  prop2
>
  .....
</elem>
```

I think this should be unified.

yanfali: This is auto formatted by prettier. The reason it does this is not introduce extra spaces which provide split points I believe.

yanfali: you added spaces, which was not the intention.

Issue #1169, opened 2022-08-24
[Bug] long form keycodes in container keys not properly converted to shorter canonical form


precondition:
### Describe the Bug

"Container keys" like LT or MT containing a long-form keycode will fail to convert the long form keycode (e.g. `KC_COMMA`) into the shorter canonical alias (e.g. `KC_COMM`) and will consequently show `N/A`.

An example of the issue can be found in the [muini32 keymap](https://github.com/qmk/qmk_configurator/blob/a23ae8c0844915b2e5fb06fbd6bc62510107cd04/public/keymaps/m/miuni32_default.json#L9).

As we can see in the source code of `parseKeycode` in `src/store/modules/parse.js`, the conversion from long form to short form is handled before any handling of container keys is done:
```js
function parseKeycode(store, keycode, stats) {
  let metadata;

  keycode = stripANY(keycode);
  stats.count += 1;

  // Check if the keycode is long-form or alternate
  if (longFormKeycodes[keycode]) {
    keycode = longFormKeycodes[keycode];
  }

  // Check if the keycode is a complex/combo keycode ie. contains ()
  if (keycode.includes('(')) {
    // Pull the keycode and or layer from within the brackets
    let key, outerKeycode;
    let splitcode = keycode.split('(');
    let maincode = splitcode[0];
    if (longFormKeycodes[maincode]) {
      maincode = longFormKeycodes[maincode];
    }
    let internal = splitcode[1];
    internal = internal.split(')')[0];

    // check for an OSM keycode
    if (maincode === 'OSM') {
      // ok we know it's OSM
      return processOneShotMods(store, keycode);
    }

    //Check whether it is a layer switching code or combo keycode
    if (internal.includes('KC')) {
      // Layer Tap keycode
      if (maincode === 'LT') {
        return newLayerContainerKey(store, maincode, internal);
      }
 // [cut for brevity]
```

PR #1182, opened 2022-09-04
Convert longform kc in container keys

precondition:
## Description

Solves issue #1169 . Read its description for more details.


yanfali (approved):
Lgtm. We really should write some tests around this code one day 😅

noroadsleft (approved):
Looks good; nice find!

PR #1185, opened 2022-09-05
Layout reset prompt when returning from "print keymap" or "test keyboard"

### LINKED ISSUE #796: layout reset prompt when returning from "print keymap" or "test keyboard"
When using a custom layout configured in the browser (not selected from a dropdown), from the "Test keyboard" window, or "Print keymap: window; when you press "< Back", the screen goes to the main interface, but attempts to clear the layout that's currently being worked on with a prompt that reads "This will clear your keymap - are you sure you want to load default keymap?"

mateossh:
## Description

Change in `mounted()` in ControllerTop component fixes first issue - prompt "This will clear your keymap - are you sure you want to load default keymap?".

Other changes are related to bug's second part - force resetting keymap name input. My approach also relies on using `keymap.dirty` from store. I created action `updateKeymapName` that updates value in store and toggles `keymap.dirty` if value isn't empty.

If you have better idea how to fix this, I'm open to suggestions.

When I was fixing this issue I merged `realKeymapName` and `keymapName`. First one was value from store, second one was input's value. Changing one value was automatically updating another. Now `keymapName` uses computed property so it can be bound to input value and no warnings are thrown.

Fixed #796

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -210,7 +210,9 @@ export default {
   },
   async mounted() {
     await this.initializeKeyboards();
-    await this.loadDefault(true);
+    if (!this.isDirty) {
+      await this.loadDefault(true);
+    }
     await this.initTemplates();
   },
   methods: {
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -99,7 +99,7 @@ export default {
   name: 'ControllerTop',
   data: () => {
     return {
-      keymapName: '',
+      // keymapName: '',
       firstRun: true
     };
   },
@@ -117,8 +117,16 @@ export default {
     isFavoriteKeyboard() {
       return this.keyboard === this.configuratorSettings.favoriteKeyboard;
     },
-    realKeymapName() {
-      return this.$store.getters['app/keymapName'];
+    // realKeymapName() {
+    //   return this.$store.getters['app/keymapName'];
+    // },
+    keymapName: {
+      get() {
+        return this.$store.state.app.keymapName;
+      },
+      set(value) {
+        this.setKeymapName(value);
+      }
     },
     keyboard: {
       get() {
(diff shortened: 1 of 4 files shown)


yanfali:
Seems like some nice clean ups. Some of this code is old and creaky. I'll test it and get back to you.

src/components/ControllerTop.vue
```diff
@@ -117,8 +116,13 @@ export default {
     isFavoriteKeyboard() {
       return this.keyboard === this.configuratorSettings.favoriteKeyboard;
     },
-    realKeymapName() {
-      return this.$store.getters['app/keymapName'];
+    keymapName: {
+      get() {
+        return this.$store.state.app.keymapName;
```
yanfali, line 121: this isn't the same.  the getter has additional functionality
```
  keymapName: (state) => {
    return state.keymapName.replace(/\s/g, '_').toLowerCase();
  },
```

yanfali (changes requested):
One minor comment, otherwise LGTM

src/components/ControllerTop.vue
mateossh: Ah, I didn't check that when I was creating PR.

After looking at the source code again I found that `setKeymapName` mutation looks like this:
```javascript
  setKeymapName(state, _keymapName) {
    state.keymapName = _keymapName.replace(/[\s/]/g, '_').toLowerCase();
  },
```
It looks like `keymapName` getter isn't necessary, because there's already value with underscores instead of whitespaces in store.

yanfali: I think this also changes as other things modify the keymap name. Try loading a keymap from JSON.

mateossh: I loaded file generated from config.qmk.fm with modified layout and "mateossh_test_layout" keymap name. 
No additional console messages, firmware did compile, downloaded files had proper names.
Then I manually set keymap name to "mateossh test layout with spaces" in file and tried to load it again in new tab.
Whitespaces were automatically changed to underscores, firmware did compile, files had proper names.

yanfali: Did UI change with the setting automatically?

mateossh: Yes

yanfali (approved):
@noroadsleft can you give it a try

noroadsleft (approved):
Works well for me; didn't have any issues. :+1:

noroadsleft:
Thanks!

PR #1186, opened 2022-09-08
Update keycode names before setting the Tester keymap

precondition:
## Description

This one-line change solves the bug wherein the tester keymap does not respect the user's selected OS keyboard layout on initial page load.


yanfali (commented):
Shouldn't this be run on keycode store itself after configuration loads?

precondition:
> Shouldn't this be run on keycode store itself after configuration loads?

Hm, I'm not sure what you mean :confused:

yanfali:
basically, this is an ordering issue I believe. The main part of the application loads first and this is lazy loaded. So when it starts up, it should probably check the value of the keymap layout and initialize itself correctly rather than calling into keycodes and reinitializing everything from scratch.

It's sort of like you start your engine. Then you start your radio, but the radio doesn't check the state of the engine, so the radio asks the engine to restart. Probably what it should do is assume the engine is initialized correctly, check that state and init properly.

yanfali:
@precondition I don't like one store reaching into another unless it really has to, as it makes debugging this more difficult in the future and may have unexpected side effects. This works equally as well and is a bit more transparent:

```javascript
   async mounted() {
     this.createKeyListeners();
+    this.$store.commit('keycodes/updateKeycodeNames');
     await this.init();
     this.setSize(this.calculateMax(this.layout));
   },
```
You can change this in VisualTesterKeymap.vue instead. it's much more obvious you are bootstrapping when the page is loading.

yanfali (approved):
Lgtm. Thanks

PR #1190, opened 2022-09-27
[chore] Run pre-commit hooks in parallel

yanfali:


mateossh:
I spent some time investigating why pre-commit hook is taking that much time. 

I started by checking (on `master`) husky pre-commit script and `eslint src` times.

Zrzut ekranu 2022-09-28 o 11 24 57

Oh, `prettier` is executed twice. Once it's executed by script, and then by `eslint`.

So I removed ESLint prettier plugin and checked times again.

Zrzut ekranu 2022-09-28 o 11 31 43

<details>
<summary>.eslintrc.js diff</summary>

```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index d83a5bc0..68c1db88 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,6 @@ module.exports = {
   plugins: ['cypress'],
   extends: [
     'plugin:vue/recommended',
-    '@vue/prettier',
     'plugin:cypress/recommended'
   ],
   rules: {
(END)
```
</details>

ESLint time went down from +4 seconds to less than 2 seconds on my M1 MacBook Air.

This PR with removed `eslint-plugin-prettier` gives following results:
Zrzut ekranu 2022-09-28 o 11 43 32

Do you think it's safe to remove `eslint-plugin-prettier`?

yanfali:
Oh good catch. If anything I would remove parallel prettier. I think the vue version has special handling for single file components. I'm not sure regular prettier does.

yanfali:
my vague memory of why we added this feature was people were sending us PRs with unformatted code. VSC IDE that I have setup formats on save so I don't need it personally, but it's useful for the general population of less sophisticated developers. I would say let's drop the call to prettier in npm and see if this causes problems on PRs.

mateossh:
How about running prettier only on changed files? I think there should be some way to do that

yanfali:
> How about running prettier only on changed files? I think there should be some way to do that

Great idea. Apparently there's a lint staged hook

mateossh:
I installed `lint-staged` and configured it [in my fork](https://github.com/mateossh/qmk_configurator/tree/commit-hooks-linting). Should I create new PR?

yanfali:
> I installed `lint-staged` and configured it [in my fork](https://github.com/mateossh/qmk_configurator/tree/commit-hooks-linting). Should I create new PR?

Sure.

PR #1197, opened 2022-10-18
Lint only changed files

mateossh:
## Description

Another approach to #1190.

I cherry-picked @yanfali commits from his PR, removed `eslint-*-prettier` packages since they won't be necessary and installed `lint-staged` package.

Questions:
1. Is having `concurrently` task necessary in `package.json`? I don't think anyone would run this manually.
2. This PR is configured to use `prettier`, not `parallel-prettier` packaged. Should I remove `@mixer/parallel-prettier` or use it instead of normal `prettier`?


yanfali:
Are they equivalent. Does prettier format vue sfc? Also parallel is much faster than regular iirc on multicore systems. Might need to verify that. Benchmarks as to why to make this change? My biggest concern is sfc. If they aren't getting formatted this isn't a replacement.

yanfali:
All the white space changes and double quotes are out of scope

mateossh:
I know. I will use this patch to benchmark `prettier`/`pprettier` performance and then remove this from PR.

yanfali:
bump

PR #1204, opened 2022-11-09
Add Missing Keyboard Settings keycodes

noroadsleft:
## Description

Adds the following keycodes to QMK Configurator:

- `CL_TOGG`: Toggle Caps Lock and Left Control swap
- `GUI_TOG`: Toggles the status of the GUI keys
- `NK_TOGG`: Toggle N-key rollover
- `AG_TOGG`: Toggle Alt and GUI swap on both sides
- `CG_TOGG`: Toggle Control and GUI swap on both sides
- `EC_SWAP`: Swap Caps Lock and Escape
- `EC_TOGG`: Toggle Caps Lock and Escape swap
- `EC_NORM`: Unswap Caps Lock and Escape
- `EH_LEFT`: Set the master half of a split keyboard as the left hand (for EE_HANDS)
- `EH_RGHT`: Set the master half of a split keyboard as the right hand (for EE_HANDS)

### DIFF
--- a/src/store/modules/keycodes/kb-settings.js
+++ b/src/store/modules/keycodes/kb-settings.js
@@ -76,6 +76,76 @@ export default [
     width: 1500
   },
   {
+    name: 'Swap Caps/Esc',
+    code: 'EC_SWAP',
+    title: 'Swap Caps Lock and Escape',
+    width: 1500
+  },
+  {
+    name: 'Split Master is Left Hand',
+    code: 'EH_LEFT',
+    title: 'Set the master half of a split keyboard as the left hand (for EE_HANDS)',
+    width: 2000
+  },
+  {
+    name: 'Split Master is Right Hand',
+    code: 'EH_RGHT',
+    title: 'Set the master half of a split keyboard as the right hand (for EE_HANDS)',
+    width: 2000
+  },
+  {
+    width: 0
+  },
+  {
+    name: 'Toggle LCtl/Caps',
+    code: 'CL_TOGG',
+    title: 'Toggle Caps Lock and Left Control swap',
+    width: 1500
+  },
+  {
+    width: 3000
+  },
+  {
+    width: 3000
+  },
+  {
+    width: 1500
+  },
+  {
+    name: 'Toggle GUI',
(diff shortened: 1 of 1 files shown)


noroadsleft:
Went for an Enable-Toggle-Disable layout, by rows:

src/store/modules/keycodes/kb-settings.js
```diff
@@ -75,6 +75,76 @@ export default [
...
+  },
+  {
+    width: 3000
+  },
+  {
+    width: 3000
+  },
+  {
+    width: 1500
+  },
```
noroadsleft, line 113: Not sure how I feel about this bit...

yanfali: You could create an alias to an object like so:
```
const blank3000Row = { width: 3000 };
```
and then use it like so:
```javascript
...bank3000Row
```

src/store/modules/keycodes/kb-settings.js
```diff
@@ -75,6 +75,76 @@ export default [
     title: 'Swap Ctrl and GUI on both sides (for macOS)',
     width: 1500
   },
+  {
+    name: 'Swap Caps/Esc',
+    code: 'EC_SWAP',
+    title: 'Swap Caps Lock and Escape',
+    width: 1500
```
yanfali, line 82: if we're using the same width in a lot of places, you can replace this with a constant.

src/store/modules/keycodes/kb-settings.js
noroadsleft: The issue I'm having with this is that I need a space that's 7.5 key units, but there isn't a spacer class in the CSS for one.

I could add one of course, but I don't think that's a scalable solution.

yanfali (approved):

yanfali:
@noroadsleft merge if you're happy with this

fauxpark:
Looks like these have all been added now, except for `EH_LEFT` and `EH_RGHT`.

yanfali:
> Looks like these have all been added now, except for `EH_LEFT` and `EH_RGHT`.

then you should probably close this PR.

PR #1207, opened 2022-11-16
Add Chinese HK and Chinese TW translation

zxx-457:


yanfali:
Hi. Can you add a read me about what the script and the data file are for? And where did it come from? I'm reluctant to add a 300k file unless I know where it's from. Thanks

zxx-457:
The large file just words mapping

package.json
```diff
@@ -31,6 +31,7 @@
     "howler": "^2.1.3",
     "keypress.js": "^2.1.5",
     "lodash": "4.17.21",
+    "simplify-traditional": "^1.0.0",
```
yanfali, line 34: move this to dev dep please. it's not required for runtime.

yanfali:
is there anyway to pull in the translations file dynamically from the original repo as a dep?

package.json
zxx-457: Sure , I,m sorry

I will write a readme

zxx-457:
> is there anyway to pull in the translations file dynamically from the original repo as a dep?

Yes it could pull the dict from the repo , and dynamic translate , but I think this Kanji map doesn't change much.

yanfali:
> > is there anyway to pull in the translations file dynamically from the original repo as a dep?
> 
> Yes it could pull the dict from the repo , and dynamic translate.
> 
> I will update the translate repo later for dynamic, thank you

Thank you. I just don't want to bloat up the repo with a file which is only used infrequently. Thanks!

zxx-457:
> > > is there anyway to pull in the translations file dynamically from the original repo as a dep?
> > 
> > 
> > Yes it could pull the dict from the repo , and dynamic translate.
> > I will update the translate repo later for dynamic, thank you
> 
> Thank you. I just don't want to bloat up the repo with a file which is only used infrequently. Thanks!

Ok

zxx-457:
Should I just add this two translates, and remove the script?

yanfali:
> Should I just add this two translates, and remove the script?

Do you think we should run the script as part of the build process?

zxx-457:
> > Should I just add this two translates, and remove the script?
> 
> Do you think we should run the script as part of the build process?

No, only use it when add simplified Chinese translates

yanfali:
> > > Should I just add this two translates, and remove the script?
> > 
> > 
> > Do you think we should run the script as part of the build process?
> 
> No, only use it when add simplified Chinese translates

I'm ok adding the script to the project. I'll let you choose. Thank you for this contribution!

zxx-457:
> > > > Should I just add this two translates, and remove the script?
> > > 
> > > 
> > > Do you think we should run the script as part of the build process?
> > 
> > 
> > No, only use it when add simplified Chinese translates
> 
> I'm ok adding the script to the project. I'll let you choose. Thank you for this contribution!

I think we could add it .

Thank you sir , I would like to contribute more If I'm capable.

yanfali:
OK. I'll check it out locally tomorrow, and if it looks good I'll merge it.

yanfali:
@noroadsleft please take a look

noroadsleft (approved):
Looks good to me.

yanfali (approved):
Thank you.

PR #1231, opened 2022-12-20
Fix Q8 ISO keymap

fauxpark:
## Description

Copied from ANSI keymap (which BTW has some macros that don't appear to be used?)

image


noroadsleft:
Thanks!

PR #1232, opened 2022-12-23
Add Caps Word Toggle keycode to Quantum section

an1z:
Adds Caps Word Toggle keycode (CW_TOGG) to QMK Configurator

### DIFF
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -427,8 +427,14 @@ export default [
     title: 'Right Alt when held, ) when tapped'
   },
   {
-    name: 'RS / Enter',
-    code: 'SC_SENT',
-    title: 'Right Shift when held, Enter when tapped'
+    name: 'rs / enter',
+    code: 'sc_sent',
+    title: 'right shift when held, enter when tapped'
+  },
+  {
+    name: 'Toggle CW',
+    code: 'CW_TOGG',
+    title: 'Toggle Caps Word (Enable Caps Lock for the next word)',
+    width: 1500
   }
 ];
--- a/src/store/modules/keycodes/quantum.js
+++ b/src/store/modules/keycodes/quantum.js
@@ -429,7 +429,7 @@ export default [
   {
     name: 'rs / enter',
     code: 'sc_sent',
-    title: 'right shift when held, enter when tapped'
+    title: 'Right shift when held, enter when tapped'
   },
   {
     name: 'Toggle CW',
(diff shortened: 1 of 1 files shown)


src/store/modules/keycodes/quantum.js
```diff
@@ -427,8 +427,14 @@ export default [
     title: 'Right Alt when held, ) when tapped'
   },
   {
-    name: 'RS / Enter',
-    code: 'SC_SENT',
-    title: 'Right Shift when held, Enter when tapped'
+    name: 'rs / enter',
+    code: 'sc_sent',
```
yanfali, line 431: Case changes?

an1z: Thanks for catching. Reverted.

yanfali:
The only issue with this is the keyboard firmware needs to enable it. CAPS_WORD_ENABLE = yes. We don't currently support that in configurator.

an1z:
> The only issue with this is the keyboard firmware needs to enable it. CAPS_WORD_ENABLE = yes. We don't currently support that in configurator.

Agree with this concern. I think we need to do the same for the volume control keys as well by adding EXTRAKEY_ENABLE = yes. I still appreciate us supporting the volume keys here for the below reason.

I feel it is still useful to support this in the configurator since the keymaps array doesn't need to be manually edited if someone wants to bind a key to the toggle caps word feature.

Looking forward to hear what you think :)

yanfali:
I'm ok with adding code to UI. @noroadsleft any other concerns?

noroadsleft:
I tested this locally and threw a `CW_TOGG` on a keymap for a board that doesn't enable the feature, and the firmware still compiled; seems QMK Firmware takes care of this on its end - probably just comes out as a NOOP.

src/store/modules/keycodes/quantum.js
```diff
@@ -430,5 +430,11 @@ export default [
     name: 'RS / Enter',
     code: 'SC_SENT',
     title: 'Right Shift when held, Enter when tapped'
+  },
+  {
+    name: 'Toggle CW',
```
noroadsleft, line 435: `CW Toggle` is more consistent with how similar keycodes are displayed. (`BL Toggle` for Backlight, instead of `Toggle BL`.)

```suggestion
    name: 'CW Toggle',
```

noroadsleft (changes requested):

src/store/modules/keycodes/quantum.js
an1z: > 

Sounds good. Updated.

noroadsleft (approved):

noroadsleft:
Thanks!

PR #1241, opened 2023-02-20
Basketweave piit79

totolouis:
Add piit79 layout.
From his website :
"Modified by [piit79](https://github.com/piit79) of 42. Keebs. This is revision 1.9 which features the following modifications:

Symmetric 3-key macro cluster on the right
Support for two rotary encoders, one on each side, either in the top or the bottom macro position. A standard switch can also be used instead of each encoder
Split backspace support"

## Description

Just added his layout from his github.


zvecr:
This wouldnt do anything as `piit79` isnt a thing within `qmk_firmware`.

PR #1246, opened 2023-02-28
Fix some more keycodes

fauxpark:


yanfali (approved):
Do these codes need to be changed in keycode map too? Or will they just be ANY?

fauxpark:
They weren't there to begin with.

yanfali:
> They weren't there to begin with.

you should try and load them into the configurator. I suspect they will be set to undefined. Iirc the parser attempts to validate the codes before using them and so it will ignore codes it doesn't have.

fauxpark:
Seems to use `ANY()` both before and after.

PR #1257, opened 2023-05-15
force `keyboardPath` to use lowercase folder names

noroadsleft:
## Description

A long time ago, I worked out a method for previewing Configurator keymaps against still-pending `info.json` updates to `qmk/qmk_firmware`, which worked great when I was on Windows, which is case-insensitive. But now I've switched to Linux Mint as my main OS, and my method no longer works. :rofl:

### DIFF
--- a/src/store/modules/app/actions.js
+++ b/src/store/modules/app/actions.js
@@ -31,7 +31,7 @@ const actions = {
    * load the default keymap for the currently selected keyboard
    */
   async loadDefaultKeymap({ state }) {
-    const keyboardPath = state.keyboard.slice(0, 1);
+    const keyboardPath = state.keyboard.slice(0, 1).toLowerCase();
     // eslint-disable-next-line
     const keyboardName = state.keyboard.replace(/\//g, '_');
     const resp = await axios.get(


noroadsleft:
yanfali [writes](https://github.com/qmk/qmk_configurator/pull/1257#discussion_r1193903532):
> Can you give examples in the pr as to what trick you are using please? That way it's in the history

For testing pending updates, one must have the Configurator development environment set up, following [the instructions on the main readme](https://github.com/qmk/qmk_configurator/blob/fcc13beac9d39e2fe9b6f74f435989e11edce13a/README.md?plain=1#L17-L70).

Start the development environment using `yarn run dev`, then access the local Configurator through the `localhost` URL it gives you when start-up completes.

Copy a keymap **written for the pending `info.json` file** to `public/keymaps/p/Preview info.json_default.json` (note the space in the target file name).

Import the pending `info.json` file using the `Ctrl+Shift+I` method.

Click the `Load Default` button.

----

Clicking the `Load Default` button here results in an [HTTP 404 Error](https://github.com/qmk/qmk_configurator/assets/18669334/7de0a8ae-7e2e-4f16-9bfd-2ad7a7f7c379) if this is attempted on a case-sensitive file system, because `p` is not the same as `P`. On Windows this works because Windows file systems are case-insensitive, so there's no difference whether the directory is named `p` or `P`. However on Linux, the system will attempt to load from a directory named (uppercase) `P` - which doesn't work because the keymap folders are lowercase in the Git repository, so one would need to create a directory `P` to accept the preview keymap, alongside the already-present `p` directory.

PR #1258, opened 2023-05-19
Expand functionality of "show key sizes" switch

fauxpark:
## Description

Replaced with a dropdown to choose one of the following (default being Keymap):
image
Matrix shows the row/col of each key, or "?" if there is none set (should be fairly uncommon now that layout macros have been moved to info.json, but may crop up when using the preview):
image
Index displays the key ordering, which should again help when verifying info.json files.
image
Size performs the existing "show key sizes" functionality.

Lastly, the parameter boxes for things like layer keys etc. are now hidden when this setting is changed away from Keymap.


src/components/SettingsPanel.vue
```diff
@@ -23,21 +23,19 @@
...
-        <toggle-button
-          id="setting-toggle-display-size"
-          :value="displaySizes"
-          :width="defaultWidth"
-          :sync="true"
-          :labels="labels"
-          @change="toggleDisplaySizes"
-        />
+        <select id="setting-panel-legends" v-model="keyLegends">
+          <option value="keymap">Keymap</option>
```
yanfali, line 34: these labels need to go into i18n too. Key them off the values. Right now the UI is the where the values are coming from. I would prefer the values come out of the store rather than be hard coded.

yanfali (commented):
Looks pretty good overall. I would just move the values into the store and generate the drop down from that. I would guess the appropriate place would be app.

src/store/modules/keymap.js
```diff
@@ -442,8 +442,8 @@ const mutations = {
       state.colorwayIndex = index;
     }
   },
-  toggleDisplaySizes(state) {
-    state.displaySizes = !state.displaySizes;
+  setLegends(state, legends) {
```
yanfali, line 445: ok, since you're using keymap to store this value, you might as well use keymap to store the legends. I think it's probably better if this were in app tho, since it's not really keymap related, more UI and app. I would validate the values here, something simple like 
```['matrix', 'size','index','keymap'].includes(legends);``` If you keep the values in the keymap store and use them to generate the UI, the UI should never be wrong.

File: src/components/BaseKey.vue [hunk 3]
```diff
@@ -134,12 +135,20 @@ export default {
...
     displayName() {
-      if (this.displaySizes) {
-        return this.uh > this.uw
-          ? this.uw === 1
-            ? this.uh
-            : `${this.uw} /\n ${this.uh}`
-          : this.uw;
+      switch (this.legends) {
+        case 'size':
+          return this.uh > this.uw
```
yanfali, line 140: let's clean this up a bit:
```javascript
const { uh, uw } = this;
```
remove all the this references, it's just noise.

yanfali:
I would consider adding a helper to BaseKey called isLegendKeymap
```javascript
/**
 * are we showing keymap legends?
 */
function showingKeymap() {
  return this.legends === 'keymap';
}
```
Then you can use that instead of all the string comparisons all over the place. That way if things change in the future you aren't touching N files.

yanfali:
I'll take a look again today. I'm traveling.

yanfali (approved):
LGTM

PR #1267, opened 2023-06-07
Fix LT functions by removing space

joe-scotto:
## Description
`keyboards/scottokeebs/scotto36` needed the space between the comma and keycode removed in order for the configurator to pick up on it. `LT(1, KC_TAB)` needed to be `LT(1, KC_TAB)` along with other `LT()` functions.

### DIFF
--- a/public/keymaps/h/handwired_scottokeebs_scotto36_default.json
+++ b/public/keymaps/h/handwired_scottokeebs_scotto36_default.json
@@ -5,10 +5,10 @@
     "layout": "LAYOUT_split_3x5_3",
     "layers": [
         [
-            "KC_Q",         "KC_W", "KC_E",    "KC_R",    "KC_T",           "KC_Y",          "KC_U",          "KC_I",     "KC_O",   "KC_P",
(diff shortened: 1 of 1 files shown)


noroadsleft:
@joe-scotto If you just let Configurator load the existing default, compile and flash, do the `LT()` keycodes work as intended on tap, or not? From what I can tell they load correctly, but the UI doesn't draw the tap keycode, even though it's loaded. But I can't test because I don't have the board.

joe-scotto:
@noroadsleft Just tested both the live site and local copy...

**Live site: `LT(1, KC_TAB)`**
The layer works when holding but the tap function does not.

**Local site: `LT(1,KC_TAB)`**
Both functions work as intended.

Seems that having a space breaks the keycode tap part of `LT()`

noroadsleft:
Thanks. I should see about making the import function smarter.

noroadsleft (approved):

yanfali:
> Thanks. I should see about making the import function smarter.

A simple replace Regexp looking for \s globally should do it.

PR #1272, opened 2023-06-13
Add ScottoNum default keymap

joe-scotto:
## Description
Add default keymap for the ScottoNum handwired macropad.


noroadsleft (approved):
Thanks!

PR #1285, opened 2023-07-07
add dp3000 macropad default keymap

depermana12:
## Description
as title describe, please add dp3000 default keymap


public/keymaps/d/dp3000_default.json
```diff
@@ -0,0 +1,57 @@
...
+      "MO(1)",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO",
+      "KC_NO"
+    ]
```
noroadsleft, line 54: This works, but it doesn't match the key assignment given when compiling from the `qmk_firmware` repository, which means users will experience different behaviour depending on how they compile their keymaps.

https://github.com/qmk/qmk_firmware/blob/632de46b006a03ff04095a00ab21179d9510e39a/keyboards/dp3000/keymaps/default/keymap.c#L35-L49

depermana12: im sorry for the inconsistent key assignment, i have resolved the issue by making the necessary changes to match the original `keymap.c` from the qmk_repository

noroadsleft (approved):
Looks good, thanks!

PR #1310, opened 2023-09-21
Add ScottoMacrodeck default keymap

joe-scotto:
## Description
Add default keymap for the ScottoMacrodeck handwired macropad.


noroadsleft (approved):
Thanks!

PR #1319, opened 2023-10-28
fix: restore QMK icon

yanfali:
- the build process wasn't able to resolve the css path for the build icon. The quickest fix was to just move the icon file to public, this ensures it's always copied when making the dist.
 - fixed some warnings in App.vue about ordering of tags
 - moved the relevant CSS in App.vue
 - removed unused file
 - updated caniuse db


noroadsleft (approved):
Thanks for this. :)

PR #1320, opened 2023-10-28
chore: clean up accumulated vue warnings

yanfali:

### DIFF
--- a/src/components/Keycode.vue
+++ b/src/components/Keycode.vue
@@ -25,16 +25,47 @@ import isUndefined from 'lodash/isUndefined';
 
 const debug = false;
 export default {
-  name: 'keycode',
+  name: 'KeycodeComponent',
   props: {
-    type: String,
-    code: String,
-    title: String,
-    width: null,
-    name: String,
-    classes: String,
-    styles: Object,
-    layer: Number
+    type: {
+      type: String,
+      default: undefined
+    },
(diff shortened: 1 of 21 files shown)


src/components/Keycode.vue
```diff
@@ -25,16 +25,47 @@ import isUndefined from 'lodash/isUndefined';
...
+      default: ''
+    },
+    title: {
+      type: String,
+      default: ''
+    },
+    width: {
+      type: String,
+      default: undefined
+    },
```
noroadsleft, line 45: This was `width: null` previously - is that no longer valid?

yanfali:
On Sat, Oct 28, 2023, 16:56 James Young ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/components/Keycode.vue
> <https://github.com/qmk/qmk_configurator/pull/1320#discussion_r1375336287>
> :
>
> > +    width: {
> +      type: String,
> +      default: undefined
> +    },
>
> This was width: null previously - is that no longer valid?
>

Iirc we check if this is undefined before using it. Can you verify that?

src/components/Keycode.vue
yanfali: on L75 we check if it's undefined, so this is equivalent.

PR #1367, opened 2024-10-23
chore: use pinia for status store

yanfali:
## Description

 - port away from vuex to pinia
 - using the options api for now until everything is ported over
 - preparation for a vue3 port
 - fixed broken workflow so it passes again


noroadsleft (approved):
Everything I can think to check still works.

Feel free to merge this if you're satisfied with current state.

PR #1369, opened 2024-11-01
Remove axios

yanfali:
## Description
Remove Axios dependency and replace with standard baseline fetch.


src/components/VisualKeymap.vue
```diff
@@ -109,8 +112,7 @@ export default {
       'resizeConfig',
       'setLoadingKeymapPromise'
     ]),
-    ...mapMutations('status', ['append']),
-    ...mapActions('status', ['scrollToEnd']),
+    ...pinia.mapActions(useStatusStore, ['append', 'scrollToEnd']),
```
yanfali, line 115: I missed these setters and getters in the last PR. So this fixes that.

src/store/modules/keymap.js
```diff
@@ -402,6 +402,9 @@ const mutations = {
   },
   initKeymap(state, { layout, layer, code = 'KC_NO' }) {
     const { name } = this.getters['keycodes/lookupKeycode'](code);
+    if (!layout) {
```
yanfali, line 405: prevents a crash on initial load because layout is undefined

PR #1370, opened 2024-11-05
Convert vuex/keycodes store to pinia

yanfali:
## Description
This PR replaces all the uses of vuex keycodes store with a pinia equivalent. It also fixes a bug in the steno code I discovered while refactoring, where steno boards were not being identified correctly.

This should be 100% equivalent with no regressions.

Vuex is in deep maintenance mode and pinia is the recommended library for vue. This work is preparation for moving to vue 3 and vue 2 is now EOL.


yanfali:
@precondition if you get a chance, a quick review or running it locally to spot if I broke anything would be appreciated. Thanks

yanfali:
I added some jsdoc for some of the hairy parts, and documented some of the internal types used inside the keycode JSON.

precondition:
Oh shoot, I planned to take a look at this later today.

yanfali:
> Oh shoot, I planned to take a look at this later today.

No worries. Any feedback you have is appreciated. Please check I didn't break anything on the deployed code. Thank you!

src/store/keycodes.js
```diff
@@ -0,0 +1,218 @@
...
+      if (this.searchFilter !== '') {
+        this.searchCounters = {
+          ANSI: countMatches(this.searchFilter, ansi),
+          'ISO/JIS': countMatches(this.searchFilter, iso_jis),
+          Quantum: countMatches(this.searchFilter, quantum),
+          KeyboardSettings: countMatches(this.searchFilter, settings),
+          AppMediaMouse: countMatches(this.searchFilter, media)
+        };
+      }
+    }
```
precondition, line 171: This is a pre-existing issue but the collections passed to `countMatches` (`ansi`, `iso_jis`, etc.) are read-only variables imported from `./modules/keycodes/###`. Consequently, these keycode collections are not localized at all causing incorrect match counts to be reported.

For example, if I search for "ß" with the German OS layout, the `ß` key will be highlighted but the count in both ANSI and ISO tabs will be "(0)". Same story if I search for "Y" with the German OS layout; DE_Y and KC_Y (=DE_Z) are both highlighted but the ANSI tab will only report a single match because it successfully finds a "Y" in the uppercase version of `KC_Y`'s `name` but `KC_Z` aka `DE_Y` has "Z" as its `name` in the `ansi` variable so the `countMatches` does not count it.

The solution would be to map the `toLocalKeycode` on each collection before passing it to the `countMatches`. I'll open a separate issue and open a PR for a possible implementation.

src/store/keycodes.js
```diff
@@ -0,0 +1,218 @@
...
+ * @typedef {Object} WidthPlaceholder - used for spacing
+ * @property {number} width - width in Key Units * 1000. e.g. 1U = 1000, 2U = 2000.
+ */
+
+/**
+ * @typedef {Object} KeycodeDefinition - metadata about a keycode
+ * @property {string} name - UI display label for the keycode
+ * @property {string} code - QMK keycode defintion
+ * @property {string} [keys] - javascript keypress id. Used by keyboard handler
+ * @property {number} [width] - width in Key Units * 1000. e.g. 1U = 1000, 2U = 2000.
```
precondition, line 194:
```suggestion
 * @typedef {Object} WidthPlaceholder - used for spacing
 * @property {number} width - width in Key Units * 1000. e.g. 1U = 1000, 2U = 2000
 */

/**
 * @typedef {Object} KeycodeDefinition - metadata about a keycode
 * @property {string} name - UI display label for the keycode
 * @property {string} code - QMK keycode definition
 * @property {string} [keys] - javascript keypress id. Used by keyboard handler
 * @property {number} [width] - width in Key Units * 1000. e.g. 1U = 1000, 2U = 2000
 ```
 Typo + inconsistent use of trailing period

File: src/store/keycodes.js [hunk 3]
```diff
@@ -0,0 +1,218 @@
...
+ *  & Record<'KeyboardSettings', number>
+ *  & Record<'AppMediaMouse', number>
+ * } SearchCounters
+ *
+ */
+
+/**
+ * @typedef {Object} KeycodeStoreState
+ * @property {Array.<KeycodeDefinition|KeycodeLabel|WidthPlaceholder>} keycodes - active keycodes
+ * @property {string} searchFilter - currently search filter
```
precondition, line 214:
```suggestion
 * @property {string} searchFilter - current query in the keycode picker search bar
````

File: src/store/modules/app/actions.js [hunk 4]
```diff
@@ -147,11 +155,12 @@ const actions = {
   },
   async changeOSKeyboardLayout({ dispatch, state, commit }, osLayout) {
     commit('setOSKeyboardLayout', osLayout);
-    // Important to call keycodes/updateKeycodeNames *before* keymap/updateKeycodeNames.
-    this.commit('keycodes/updateKeycodeNames');
+    const keycodesStore = useKeycodesStore();
+
+    // Important to call keycodes/updatekeycodeNames *before* keymap/updateKeycodeNames.
```
precondition, line 160:
```suggestion
    // Important to call keycodesStore.updateKeycodeNames *before* keymap/updateKeycodeNames.
```
Updating the comment to reflect changes to the call signature

precondition (commented):
I pointed out a few typos but I did not notice any regressions. However, during testing, I did uncover a pre-existing bug in the match count feature.

src/store/keycodes.js
yanfali: Good catch.

Issue #1363, opened 2024-09-20
[Bug] Printing Tool Displays Unintuitive Key Labels Instead of Readable Key Names


jeffersonjpr:
### Describe the Bug

Default tool key labels:
![image](https://github.com/user-attachments/assets/89676964-bf50-48bc-99b6-7826b00f2710)

Example from Chrome's Print Feature:
![image](https://github.com/user-attachments/assets/ba7fcbc1-11f8-4c3f-ab5f-3d647b7da594)

QMK Configurator print feature:
![image](https://github.com/user-attachments/assets/3d9285f9-032a-43e0-a284-030c80a96915)

yanfali:
Sorry. Could you give more context. Which one is your preference? Unintuitive is somewhat subjective so it's difficult to understand. Maybe some annotation around which representation you think is problematic and some possible solutions.

jeffersonjpr:
I apologize if my explanation wasn't clear. Please take a look at these two keys:

![image](https://github.com/user-attachments/assets/546a5345-4cd8-453f-ab96-d448913eed3d)
The behavior is: if I hold the key, it activates layer 2; if I just tap it, it triggers other keys.

Notice the difference in how the key behavior is represented in the printing tool. It shows the enum definitions from the code rather than the actual symbols they represent.
![image](https://github.com/user-attachments/assets/25da560a-8ff1-4830-ada2-f4cae01b274f)
The original representation in the configurator is just fine. I think the layout should remain the same in the printing tool.

yanfali:
Interesting that it's doing that. Maybe it should be a toggle.

yanfali:
Took some time to look at this today and the issue is how they keys are represented in the printing view. In the visual keymap there are specialized components for each of these container types. In the PrintKey.vue component we just have the single component and it tries it's best to do the right thing for each keycode.

One quick fix, would be to stop printing the QMK code and just show the display version.

![image](https://github.com/user-attachments/assets/356f90e4-a38a-451f-871d-238a1e0b1725)

If that's acceptable I'll open a PR.

yanfali:
ok, this should be resolved. I'll close it for now. Thanks.

PR #1371, opened 2024-11-07
fix: show the name instead of the code in print view

yanfali:
Issue [1363](https://github.com/qmk/qmk_configurator/issues/1363)

Change the PrintKey to show the `.name` instead of the `.code` attribute because it's less human friendly. Remove some of the extraneous formatting around the element.

This stems slightly from the different way we render Key components in the printing vs editing.


yanfali:
## Before
image

## After

image

noroadsleft (approved):
Looks good to me.

PR #1382, opened 2024-12-20
Slight adjustment to keycode picker for dark mode

fauxpark:
## Description

Lightened the dark mode background colour slightly, and increased the border radius by 1px, since (at least to me) they just look like solid black squares. I didn't even realise there was a border on them since I almost never use light mode.

Before:
image
After:
image


yanfali (approved):
lgtm

PR #1392, opened 2025-01-14
Move lighting keycodes to new tab

fauxpark:
## Description

Added a new Lighting tab to the keycode picker.
image

- Updated `RGB_` keycodes to `UG_`
- Removed deprecated RGB effect keycodes
- Added LED & RGB Matrix keycode sets


src/store/modules/keycodes/lighting.js
noroadsleft, line 1: It would be nice if the labels (the `name` fields) had something to differentiate them from each other. Visually, all the Brightness Up keys look the same visually, and you have to hover with the mouse cursor to tell what they are when they're assigned in a keymap.

That's my only complaint about this.

fauxpark: Yeah, it is a little annoying, but I'm not sure how to pack that much info into the legends in a nice enough way, and since these keys are typically bunched together into sensible arrangements (at least when they're all on the same layer - the same arrangement on different layers is another story) I felt it was a reasonable compromise. I'll take another stab at it though.

fauxpark: image
Here's what it looks like with prefixes added - I think it's kinda cramped, but perhaps we could increase the key size a little?

PR #1403, opened 2025-06-20
Clean up KB settings and app/media/mouse keycode tabs

fauxpark:
## Description

Keyboard settings tab:
 - Keycodes are now grouped better, and more consistently named & described. Notably, there is a gap as the `GE_TOGG` keycode does not exist.

image

App/Media/Mouse tab:
 - Tweaked keycode spacing and ordering, in particular the multimedia keys, to make a little more intuitive sense (eg. placing prev/rewind/ffwd/next at either end of the play/pause and stop keys). The Application section could perhaps use some more refinement, but the height of the tab content area is currently hardcoded to 412px due to how the keycode elements are laid out, and adding more sections would overflow it.

image

### DIFF
--- a/src/store/modules/keycodes/kb-settings.js
+++ b/src/store/modules/keycodes/kb-settings.js
@@ -1,32 +1,80 @@
 export default [
   { label: 'KeyboardSettings', group: true },
 
-  { label: 'Keyboard settings (persistent)', width: 'label' },
+  { label: 'Modifier Swapping', width: 'label' },
 
   {
-    name: 'Swap LCtl⇄Caps',
-    code: 'CL_SWAP',
-    title: 'Swap Left Control and Caps Lock',
+    name: 'Toggle Ctl⇄GUI',
+    code: 'CG_TOGG',
+    title: 'Toggle swapping both Control and GUI keys',
     width: 1500
   },
   {
-    name: 'Caps=LCtl',
-    code: 'CL_CTRL',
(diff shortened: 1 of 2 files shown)


src/store/modules/keycodes/kb-settings.js
```diff
@@ -1,159 +1,230 @@
...
+    name: 'Swap \\⇄Bspc',
+    code: 'BS_SWAP',
+    title: 'Swap Backslash and Backspace',
     width: 1500
   },
   {
-    name: 'Caps≠LCtl',
-    code: 'CL_CAPS',
-    title: 'Stop treating Caps Lock as Left Control',
+    name: 'Unswap \\⇄Bspc',
```
yanfali, line 121: you might want to use the unicode escaped versions of characters instead, but otherwise lgtm

yanfali (approved):

src/store/modules/keycodes/kb-settings.js
fauxpark: Precedent seems to be to not escape Unicode characters, for example `KC_TRNS`, but I can fix them all up in a separate PR if you want.

yanfali: Encoded characters can be harder to edit and ambiguous. It's not super critical. Folks are putting emoji in code these days so it's the practices have changed.

PR #1404, opened 2025-06-21
Improve keycode picker container keys

fauxpark:
## Description

- The `Any` key now has an "input box" below it, to match the other container-type keys eg. the Layer- and Mod-Taps, both in the picker and the keymap itself
- Widened said input box and added an ellipsis, to further reinforce its purpose

image


yanfali (approved):
lgtm

PR #1407, opened 2025-06-22
Keycode picker: rework Quantum tab

fauxpark:
## Description

image

 - `QK_REP` and `QK_AREP` moved down to Special action keys section
 - Added `QK_REBT` and `PDF()`
 - Renamed `ALL_T` -> `HYPR_T`
 - Renamed `SGUI` -> `LSG`
 - Adjusted descriptions for a bunch of keys to more closely match docs
 - Reorganised mods/mod-tap section - ~~a few gaps here where there are some missing keycodes in core~~ now fixed:

```
LCS
LCS_T <-- C_S_T
LCG
LCG_T
LCSG
LCSG_T
LSAG
LSAG_T

RCA
RCA_T
RCG
RCG_T
RCSG
RCSG_T
RCAG
RSAG
RSAG_T
```

PR #1408, opened 2025-06-29
Keycode picker: add missing mod & mod-tap keys

fauxpark:
## Description

Following up #1407, now that we have the full set in QMK core (https://github.com/qmk/qmk_firmware/pull/25399)

image


yanfali (approved):
Lgtm

PR #1410, opened 2025-07-07
Keycode picker: improve mod keycode generation

fauxpark:
## Description

Repurposed the `osmLookup` table to generate not just OSM keys, but mod and mod-taps as well.

Now possible after https://github.com/qmk/qmk_firmware/pull/25443.


src/store/modules/keycodes/quantum.js
```diff
@@ -10,68 +10,85 @@ function makeLT(layer) {
...
+  const [name, desc] = tuple;
+  return {
+    name: `${name}_T`,
+    code: `${code}_T(kc)`,
+    type: 'container',
+    title: `${desc} when held, kc when tapped`
+  };
+}
+
+// make a One-Shot Mod Keycode Definition
```
yanfali, line 81: Same

yanfali (commented):
Seems like a good change. Use jsdoc for functions it can also do minimal type checks in modern tooling.

yanfali (approved):

PR #1412, opened 2025-07-12
Keycode picker: add `title`s for ANSI & ISO entries

fauxpark:
## Description

These keycodes were missing their descriptions.

A good chunk of this (alphanumerics etc) will not actually do anything, however, because of the way the language-specific keycode system is implemented - I think they will need to be given descriptions over at qmk_firmware and then regenerated here.

### DIFF
--- a/src/store/modules/keycodes/ansi.js
+++ b/src/store/modules/keycodes/ansi.js
@@ -1,157 +1,303 @@
 export default [
   { label: 'ANSI', width: 'label', group: true },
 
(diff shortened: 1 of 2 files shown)


yanfali:
What does titles add? Wouldn't it fall back to name?

fauxpark:
Apparently not, for example `KC_BSPC` currently shows:
image

With this change:
image

And I tried to align the titles with the descriptions given in the keycode docs.

yanfali (approved):
Lgtm

PR #1413, opened 2025-07-13
Use `v-select` for layout dropdown

fauxpark:
## Description

For consistency.

image

### DIFF
--- a/src/components/ControllerTop.vue
+++ b/src/components/ControllerTop.vue
@@ -44,15 +44,14 @@
           class="drop-label"
           >{{ $t('layout.label') }}:</label
         >
-        <select id="layout" v-model="layout" @focus="focus" @blur="blur">
-          <option
-            v-for="(aLayout, layoutName) in layouts"
-            :key="layoutName"
-            :value="layoutName"
-          >
-            {{ layoutName }}
-          </option>
-        </select>
+        <v-select
+          v-model="layout"
+          max-height="600px"
+          :clearable="false"
+          :options="Object.keys(layouts)"
+          @search:focus="focus"
+          @search:blur="blur"
+        ></v-select>
       </div>
       <div class="topctrl-keymap-name">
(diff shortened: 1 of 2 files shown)


yanfali (commented):
Hard coding layout into the widget is an anti pattern. If you use CSS you can override. What's the motivation?

fauxpark:
If you mean the `max-height` that's just what the keyboard `v-select` does, I simply copied what it's doing 🤷

Can fix it if you want.

yanfali:
Why was the ID and layout CSS removed? Does it not do anything?

fauxpark:
It was just to make the `select` look closer to the `v-select`, it seems:
image

yanfali:
are you testing on FF? The layout looks completely broken on there.

yanfali:
Ah ok. I see what you're going for. So the only reason iirc we use v-layout is to make a searchable list. The number of layouts for your typical keyboard was small enough this wasn't considered necessary.
I guess I'm ok using it for layout, but it doesn't really need it.

fauxpark:
Yes, I was testing on Firefox but just checked Chrome as well, all looks good to me?

I can add `:searchable="false"` if you want; the main thing was that the two different dropdown styles looked kind of odd.

yanfali:
I notice the down arrows aren't aligned... 🫠

fauxpark:
Looks like the entire keyboard dropdown is shifted 1px to the left (or the other two to the right). Increasing `#drop-label-keyboard`'s `min-width` to 138px fixes it. `.drop-label` is 160px by default, less 22px for the favorite button, but I don't know how consistent that is...

fauxpark:
You can see from the before image this was already misaligned, just probably less obvious due to the different dropdown styles.

yanfali:
> You can see from the before image this was already misaligned, just probably less obvious due to the different dropdown styles.

Fair. If you can fix it that would be great.

PR #1414, opened 2025-07-19
Tidy up Settings panel

fauxpark:
## Description

|Before|After|
|------|-----|
|image|image|

The dropdowns will still be in monospace font due to a CSS rule on `select`s which would also affect the layout dropdown in `ControllerTop` - should be irrelevant and fine to remove after #1413.


yanfali:
That's a really nice improvement.

fauxpark:
image

How about this?

I intentionally removed the help text for language and dark mode as I thought they were fairly self explanatory.

yanfali:
can you add a bit more spacing between items? I think they feel a bit too bunched up still. Otherwise that's an improvement.

yanfali:
No. That's kind of ugly.  I would say keep the clusters of options together, but increase separation between the options. Think of them as tuples.

fauxpark:
light
dark

Per suggestions

OS keyboard layout dropdown is a bit of a pain since its options make it quite long. Did my best to make it look OK. They'll be slightly shorter without the monospace font.

Might look at giving the toggle switches dark mode styling, they seem too bright and the labels are hard to read in the off position.

yanfali (approved):
Nicely done.

PR #1415, opened 2025-07-20
Update InfoBar styling

fauxpark:
## Description

I thought this looked a little better.

Before
image

After
image
(colon instead of dash is from another experiment regarding language-specific keycodes)

Could also add a drop shadow, and IMO the background colour doesn't really fit with the rest of the UI, but I've left those for now.


fauxpark:
Keycode without description:
image
With long(ish) description:
image

PR #1417, opened 2025-07-20
chore: remove unused dep serialize-javascript

yanfali:
- can't find any uses for this dep
 - update caniuse


yanfali:
obsoletes #1396

fauxpark (approved):
Interesting that copilot doesn't see the yarn.lock

PR #1418, opened 2025-07-21
Fix "Open keyboard folder on GitHub" color

fauxpark:
## Description

Small visual bug, in light mode the icon is hard to see. Using `unset` since removing the `color` would make it the link color.

image
image


yanfali:
Interesting. Couldn't we just provide light and dark versions?

PR #1419, opened 2025-07-25
Potato facts improvements

fauxpark:
## Description

- Added some missing periods
- Removed enclosing ""s where not necessary
- Adjusted punctuation to match typical for each language (as best I can tell)
- Fixed translations not applying upon switching language (see below)
- Removed a vestigial pre-i18n list of facts - some not present in new list, not sure why
- Added a fade out-in transition when new fact is generated

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -3,7 +3,11 @@
     <span style="display: none">{{ revision }}</span>
     <div>
       <header>
-        <p class="random-potato">{{ $t(potatoFactId) }}</p>
+        <Transition name="potato-fade" mode="out-in">
+          <p class="random-potato" :key="potatoFactId">
+            {{ $t(potatoFactId) }}
+          </p>
+        </Transition>
       </header>
       <router-view />
       <spinner :is-visible="showSpinner" :status="spinnerMsg" />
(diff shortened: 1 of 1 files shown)


fauxpark:
These specifically or in general?

yanfali:
Like are you changing your browser language?

fauxpark:
Just changing the language in settings. If the double quotes are unbalanced or improperly escaped it fails to compile, otherwise I'm not paying too much attention since this isn't changing any actual wording.

I did find the message doesn't update immediately when changing language, perhaps using computed will fix that. Also, there is a `potato-facts.js` with all the English strings under `src/` that doesn't seem to be referenced anywhere, vestigial?

yanfali:
Try removing it and seeing what breaks 😀

fauxpark:
Nothing exploded.

Language switching fixed, and added the placeholder to i18n - I did mostly reference Google Translate as I figured it was safe enough for this very simple sentence. It seems to be flashing the English placeholder anyway though, no idea whether I did something dumb or it's just how vue-i18n works.

yanfali (approved):
Looks good overall. Changing other languages can be dodgy even the age of llm, but let's go with it and if someone complains they can open a PR or issue.

PR #1420, opened 2025-07-27
Add hover state for buttons

fauxpark:
## Description

image

Seems like `.ui-button` is intended as a common button class but it is used almost nowhere, and the styles are applied to specific IDs instead...

Also disabled selection & I-beam pointer for the "keymap.json" label.

### DIFF
--- a/src/scss/style.scss
+++ b/src/scss/style.scss
@@ -276,7 +276,6 @@ button + button {
 }
 
 .label-button {
-  pointer-events: none;
   user-select: none;
 }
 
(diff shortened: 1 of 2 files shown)


src/scss/style.scss
```diff
@@ -275,6 +275,11 @@ button + button {
   margin-left: 4px;
 }
 
+.label-button {
+  pointer-events: none;
```
yanfali, line 279: are you sure you want to disable point-events? doesn't that disable the indicator it's clickable?

PR #1422, opened 2025-07-28
Remove references to Travis in env vars

fauxpark:
## Description

Travis is long gone...

Fixed some inconsistencies as well - I think the develop deployment is still using `keyboards.qmk.fm` since the `sed` command would not match anything to replace it with `keyboards.develop.qmk.fm`.

`deploy.sh` appears to be unused since the days of Travis.


fauxpark:
Not sure what you mean. All this is doing is replacing references to `TRAVIS` -> `GHA` (purely cosmetic), and changing the `VUE_APP_` prefix to `VITE_` in `.env.production` which I guess were missed in the switch to Vite. There are no other references to `VUE_APP` and the base `.env` contains the same vars but with the `VITE_` prefix which *are* referenced in the code (for example https://github.com/qmk/qmk_configurator/blob/master/src/store/modules/constants.js#L4-L8).

This still works because in a `production` build, both `.env` files are picked up, and the `VUE_APP_` vars would have been left unused. So with this PR there is not even any effective change between `development` and `production` builds since both sets of values are identical anyway. The only thing that differs is now the `sed` command correctly adjusts the keyboards URL to the `develop` instance in the develop deploy workflow.

yanfali:
Fair enough. I don't remember making the change. It's been a minute.

yanfali (approved):
Approved

PR #1435, opened 2025-09-13
Fix for `KC_NO` not triggering the InfoBar

noroadsleft:
PR #1428 introduced a check[^1] which allows hovering a keycode in the Keycode Selection array to trigger the InfoBar only if the key being hovered has a `name` property. This in turn caused hovering the `KC_NO` keycode to no longer trigger the InfoBar, because its `name` property was an empty string.[^2]

This commit sets the `name` property of the `KC_NO` keycode to 'N/A' in order to bypass the aforementioned check, and thus trigger the InfoBar.

*Before on top, after on bottom:*
vlcsnap-2025-09-13-15h54m29s064

[^1]: https://github.com/qmk/qmk_configurator/blame/87751f5d9d93dd8995f31ac79ef601bdf50cb3bc/src/components/Keycodes.vue#L130
[^2]: https://github.com/qmk/qmk_configurator/blame/87751f5d9d93dd8995f31ac79ef601bdf50cb3bc/src/store/modules/keycodes/quantum.js#L121

### DIFF
--- a/src/scss/style.scss
+++ b/src/scss/style.scss
@@ -645,10 +645,6 @@ button {
   z-index: 1001;
 }
 
-.keycode:empty:after {
-  content: 'N/A';
-}
-
 .keycode-text,
 .keycode-container,
 .keycode-layer,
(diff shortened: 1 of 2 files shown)


fauxpark (approved):
Not sure how I missed that. There's a CSS rule on empty keycodes that adds the `N/A`; it probably should be removed to make this more obvious (plus KC_NO should be the only keycode with this name anyway).

